PR #10, opened 2024-11-08
Move remainder of CI to github

brryan:
Convert gitlab PR and nightly workflows to github.

- [x] Get something working on GPUs.
- [x] Merge branch directly into `develop` to better test the result
- [x] Fix unpredictable advection test behavior
- [x] Stop creating unique run directories; create `tst` directory in executable location
- [x] Support `--exe` flag
- [x] Support automatically finding `artemis` executable in current directory
- [x] Support default usage of building a new executable
- [x] Have the `advection` test clean up its own history files


.github/workflows/ci.yml
```diff
@@ -0,0 +1,132 @@
...
+          retention-days: 3
+      - name: Upload figures
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: figs
+          path: tst/figs
+          retention-days: 3
+
+  ci_doc_job:
```
adamdempsey90: I don't think this one is needed given `docs.yml`?

adamdempsey90 (approved):
LGTM

.github/workflows/ci.yml
brryan: Do we want to split up building and deploying docs? E.g. do we want to check that docs build for every PR, but only deploy the docs whenever something gets merged into `develop`?

adamdempsey90: That's how it's setup currently. It will only deploy on a push to develop. An MR will just build.

brryan: Ah I see it now, OK I removed all docs stuff from the CI

tst/run_tests.py
```diff
@@ -275,6 +295,13 @@ def log_init(args):
         help="do not recompile the code and reuse the build directory.",
     )
 
+    parser.add_argument(
+        "--exe",
```
brryan, line 327: @pdmullen I finally got around to making the executable an optional input argument to the test scripting, which helped with the other CI automation I'm doing. 

But what do you think about removing the ability of `run_tests.py` to build the code altogether? It drives some potential inconsistencies especially in our CI right now, with e.g. passing custom cmake flags to `run_tests.py` rather than using the existing `bash/env` script and our `CMakePresets.json`. Would requiring that the user pass an executable to `run_tests.py` break any workflows?

Note that I also changed `run_tests.py` to always run the code in a custom directory in `tst` with the date and time of when `run_tests.py` was called.

pdmullen: I think that a reason why `artemis` might be attractive to some folks is because they'll feel right at home if they have a background with `AthenaK`/`Athena++` (this notion at least partially explains some of my pushback on the "node input" direction).  

In that vein, here, the testing infrastructure was very much so derived from `AthenaK` (and `Athena++`)... I think it would be nice that the `run_tests.py` infrastructure continues to permit the opportunity to build the code for you.  It would be nice if this could directly happen with `env/bash` though, but sometimes python environments mess with this.

EDIT: To add, I kinda like that `run_tests.py` is a one stop shop to answering, e.g., did I break dust:

`python3 run_tests.py advection`

as opposed to 

`source env/bash`
`build_artemis`
`python3 run_tests.py advection --exe=now_where_did_that_exec_actually_build`

pdmullen: And thanks for adding the missing executable option!

adamdempsey90: One option is to pass just the directory where the build should live, and maybe that default to the current working directory. If `src/artemis` exists, then don't build the code, just run the test. But if `src/artemis` doesn't exist, then build and run the test. That's kind of what we already have, just without specifying `--save_build --reuse_build`. 

I would hope that as developers, we already have a directory with a working build. So just telling `run_tests.py` where that build is should be all we need. 

We should also hook this into a target for `CMake` so you can just do `make test` and that runs `run_tests.py` for you.

pdmullen: > I would hope that as developers, we already have a directory with a working build. So just telling run_tests.py where that build is should be all we need.

Everybody has different workflows 😅

brryan: @pdmullen OK if this is how you use the code now and Athena*K developers might like it as well, let's leave it as is. I never hit the rhythm of calling `run_tests.py` directly as a quick check because the build time is so long for Artemis. 

@adamdempsey90 I'm not sure we have a single favored directory for where the build should live (e.g. the bash script produces a `build_$PARTITION` directory), and so there may not be much advantage to passing a base dir rather than an executable, and maybe even a disadvantage since passing the base dir assumes that artemis then lives in `src/artemis`. 

My personal feeling is that `ctest` doesn't give us much we don't have; ultimately we're just after PASS or FAIL indicators, and we already have that. If migrating to `ctest` would help with compatibility with the rest of the Parthenon ecosystem then maybe we should get on that at some point, but in that case I think it would make the most sense to just remove our own `run_tests.py` infrastructure and replace with `ctest` entirely.

brryan: Just for context: the one other thing I want to do with this PR is improve the `run_local_ci.py` script so it will launch a batch job on chicoma for GPU testing (and maybe slow CPU testing) and report to github PRs and then I think this is ready for re-review.

pdmullen: > I never hit the rhythm of calling run_tests.py directly as a quick check because the build time is so long for Artemis.

Yeah it certainly isn't a quick check... at least we don't have to recompile for every pgen :) One plus about this entire rebuild is that it does start from a clean slate, bypassing issues one might have encountered with a `make clean` not doing what one might have thought.

> My personal feeling is that ctest doesn't give us much we don't have; ultimately we're just after PASS or FAIL indicators, and we already have that. If migrating to ctest would help with compatibility with the rest of the Parthenon ecosystem then maybe we should get on that at some point, but in that case I think it would make the most sense to just remove our own run_tests.py infrastructure and replace with ctest entirely.

I agree... although I suspect I am biased because I am more comfortable in a `python` ecosystem than `ctest`.

adamdempsey90: Not sure I follow. Our CMake build system forces artemis to live in `src/` under the build directory. All I'm suggesting is that we should be able to run the tests from any valid build directory. All that should be needed is the path to the source and a path to the executable. 

Because (as you say) our build times are so long, it's very annoying to be forced into having only one testing build that lives in `tst/build`. I want to be able to run the tests from all of my existing builds.

brryan: My concern is that, although our builds are always born in the `src/` directory, someone could copy the executable somewhere else, like a project release folder, and then we couldn't run tests on that executable if we've imposed that the executable is living in a directory called `src/`. 

I also want to be able to run the tests from all of my existing builds, and that's what this PR should accomplish. Just pass the path to whatever artemis executable youve already built that you want to test to `--exe`, and this code will now run all the tests you request on that executable without building anything. The results will also be placed into a custom `test_run_[date]_[time]` folder.

adamdempsey90: if you're going to pass a path to an arbitrary location where the binary is, then you'll also have to pass the path to the artemis source directory to know where the input files are, right? Right now it assumes that directory is one up from the binary directory (which isn't always true). Also, if you are thinking of running the tests on some detached executable that was copied to somewhere, you'll still need to have access to the original source (both knowing where it is and having read permissions). We also know that running on a mac can easily fail if you don't copy the rebound `.so` around with you. 

Alternatively (or maybe in addition to the above model), If you have access to the `CMakeCache.txt` file, you could pull out both `artemis_BINARY_DIR` and `artemis_SOURCE_DIR` to get all the paths you need. You could also pull out `ARTEMIS_ENABLE_CUDA` to know if it's a gpu build or not, `CMAKE_BUILD_TYPE`, etc. To me, that sounds ideal from a minimal amount of user action to successfully run the tests (just `python3 run_tests.py mysuite.suite` from a complete build directory).

brryan: My intention was that input files will be found in a relative location from the `run_tests.py` script so that should work regardless of the location of the executable. However, I think as written the code assumes that `run_tests.py` is being run *from* the `tst` directory -- I will fix this.

Why do we need read permissions on the original source? To run the tests one only needs read permissions on the source tree from which one is running the `run_tests.py` script, so they already have read/execute permissions, right?

I think changing the default behavior of `python run_tests.py [suite]` is out of scope for this PR -- I'm just adding an option to instead pass an arbitrary executable, which I want/need for this PR and my own preferred workflow. If we want to make `python3 run_tests.py` default to e.g. the default build directory from `env/bash` and not rebuild the executable if it exists, I think that's a separate conversation. One issue with that is that right now by default `run_tests.py` always builds a new executable so the test result is a reliable measurement of the current source tree. Having the script first look for an executable and use it if found would be a change in behavior that might be undesirable. One path forward could be to have `run_tests.py` always build in a directory consistent with `env/bash` (either e.g. `build_darwin_skylake-gold` or `build` for unknown platforms) and default to a full build, but if `--reuse_build` is provided, then use the existing executable if found in the default path. That would prevent users from accidentally testing an old executable. But again, I think this is a separate discussion that should be its own PR and discussion about desired default behavior that balances brevity of the "most common" usage with unexpected behind-the-scenes behavior.

I think the rebound mac issue is separate and right now if someone wants to move the executable on mac they need to move the rebound library as well; I think the solution here is to fix the mac rpath issue which will solve this.

One question I do think we should decide here is whether we want to change *run* behavior to always use a custom directory in `tst` (which avoids issues with e.g. `advection.dat` breaking the test framework if run multiple times), or if we should e.g. always run in the source directory by default (custom run dirs could be a runtime option to `run_tests.py`). What do you think @adamdempsey90 @pdmullen ?

pdmullen: I don't have too strong of feelings on this one @brryan.  IIRC, `advection.dat` would never break the testing framework if `run_tests.py` was invoked as originally implemented (i.e., starting from a clean build every time).  

The introduction of `--save_build` and `--reuse_build` is what brought about the conflict, and I am completely open to changing that utility.  I think your solution is an elegant one... i.e., making a custom outputs directory that contains e.g., date and time.

brryan: Yeah it's `--save_build/--reuse_build` that conflicts with the advection output. 

My only concern about the bespoke run directories is you can easily accumulate a lot of them, and while `rm -rf test_run_*` takes care of them, I wonder if we want to have some default behavior that only preserves the most recent directory or something. Designing this kind of autocleanup is tricky. We could also just only have one run directory that gets wiped between every invocation of `run_tests.py`.

pdmullen: If it were the entire build directory that was generating multiple copies, I might be more concerned.

But if it is only the output files generated by the regression that appear in unique outputs directories, that doesn't seem all too bad.  And I don't think it is too big an ask for our users to cleanup as they see fit... in fact some users may wish that we weren't calling `rm -rf` at all.  

Again, I don't feel too strongly here.

adamdempsey90: My (relatively strong) opinion is to follow the lead of other codes that have many tests and a lot of experience. For the code I'm thinking of, it's organized as

```
build/
build/path/to/exe
build/testing/
build/testing/test1
build/testing/test2
```
So every time you run `run_tests.py` from `build/` it deletes `build/testing`, remakes it, and then each test gets its own subdirectory. Also, since the tests are run from the Makefile, if the source files change, it will recompile the code and then run the tests. The code is actually run from `build/test1/` (not copying the executable, just running it), and all the outputs from that test lives in its directory. This is much cleaner than the literal soup of output files in our `build/src` directory or the soup of figure files in the `figs/` directory, currently. As we (hopefully) add more and more tests, our current model will be untenable. 

I don't think we need unique testing directories for each invocation, either. That sounds like overkill to me.

brryan: I agree about 
 - Subdividing tests into their own directories
 - Not having separate directories for each invocation

but I would like to 

 - Split this into `build` and `run` directories to accommodate the provide-your-own-executable case (so the `tst/build` directory is optional, basically). 

As to auto-rebuilding, I especially don't trust cmake to appropriately check that the configure environment is still appropriate, even if we explicitly call `cmake` every time to re-configure. For example if someone is on darwin, runs tests on skylake, then tries to run tests later on volta, I would expect cmake to fail with a bad build environment. So I lean towards the default `python run_tests.py` to always rebuild the code, and `python run_tests.py --reuse_build` to use whats in `build/path/to/exe` if it exists (and also maybe call `build_artemis -c -b build/path/to/exe` which will re-call `cmake` and `make` to do a best attempt at updating the executable if necessary). 

I've almost got this PR working with Darwin (unfortunately we can't run the CI on chicoma GPUs because they don't have internet access to report results), you just have to `ssh` into a darwin frontend and do `./tst/launch_ci_runner.py [PR number]`

adamdempsey90: I think for something like the CI or for someone running `run_tests.py` from a non-build directory, rebuilding the code by default unless `--reuse_build` is on makes sense. But, if I'm executing it from an existing build directory or passing it an exe, why would I want it to rebuild? In those scenarios I've probably already rebuilt it or at the very least I'm aware if it needs to be rebuilt or not. The default for that use case should be `--reuse_build` and I shouldn't have to specify it. 

Again, though, I think the simplest thing is to just execute this from a build directory, assume the build stage has been taken care of, and just run the tests -- as its name implies -- and nothing more.  We already have `build_artemis` or the super simple `cmake --preset=... /path/to/artemis && make -j` for the building part, so why conflate that with running the tests?

brryan: Oh I see, I missed your point about this behavior when running specifically from a `build/` directory. Then I agree, by default the existing executable should not be rebuilt, like how we don't try to rebuild an executable provided by `--exe`. But when just calling `run_tests.py` from e.g. `artemis` or `artemis/tst`, I think default behavior that builds a current executable and then runs tests on it is best to avoid users being surprised by testing the wrong thing. We can provide plenty of documentation about more efficient ways to run tests, and indeed in general I personally like to have "build" and "test" steps be separate. 

The alternative would be to completely remove the ability to build the code from `run_tests.py` but @pdmullen had fair concerns about alienating the Athena*K community with that.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+## Background
+
+## Description of Changes
+
+## Checklist
+
+- [ ] New features are documented
+- [ ] Tests added for bug fixes and new features
+- [ ] (@lanl.gov employees) Update copyright on changed files
```
brryan, line 9: @adamdempsey90 @pdmullen Any thoughts on a default PR template? I wanted to keep it simple, but background and description of changes are both quite useful IMO

pdmullen: Simple is good 😄 I also don't think we need a `CHANGELOG.md` as `parth`.

adamdempsey90: If we start putting out releases, we should have a Changelog. It doesn't have to be formatted like parthenon's though.

brryan: That's a good point about changelogs for release.  I think I would like to move away from the Parthenon approach of requiring every PR to update a living changelog, to some kind of automated system that extracts all the PRs merged into develop from the git commit history and creates a template we can use to write custom release notes, with confidence that we are covering all the major changed features. Just a thought, down the road we should iterate on that.

tst/run_tests.py
adamdempsey90: For the record, (1) I don't think we're going to alienate anyone, (2) we're not athena*, and (3) I would like for us to improve upon the testing/building infrastructure that originated from athena* as needed.

.github/PULL_REQUEST_TEMPLATE.md
adamdempsey90: We could collect changelog entries somewhere that describe things that were added, fixed, removed, etc. since the last release. And then at release time collate them into the final Changelog. 

My preference is to separate commit messages from descriptions of changes for users. The latter should be more descriptive and not include things like implementation details / source code stuff.

tst/run_tests.py
brryan: As discussed yesterday we're supporting three use cases for the test infra

.github/PULL_REQUEST_TEMPLATE.md
brryan: Yeah I agree, commit messages and changelog descriptions for users are pretty different.

tst/scripts/utils/artemis.py
```diff
@@ -20,12 +20,47 @@
...
+artemis_executable = os.path.join(artemis_dir, "tst", "build", "src", "artemis")
+artemis_inputs_dir = os.path.join(artemis_dir, "inputs")
+# artemis_fig_dir = "./figs/"
+artemis_run_dir = os.path.join(artemis_dir, "tst", "build", "src", "tst")
+artemis_fig_dir = os.path.join(artemis_dir, "tst", "figs")
+artemis_log_dir = os.path.join(artemis_dir, "tst")
+custom_exe = False
+
+# Create run directory for this invocation of the test framework
+# now = datetime.datetime.now()
```
pdmullen: Should these lines go away now?

tst/suites/gpu.suite
```diff
@@ -19,4 +19,4 @@ binary/binary
 binary_adi/binary_adi
 nbody/nbody
 diffusion/viscous_diffusion
-diffusion/thermal_diffusion
+#diffusion/thermal_diffusion
```
pdmullen, line 22: Time constraint?

tst/suites/parallel.suite
```diff
@@ -14,14 +14,11 @@
 # parallel suite
 
 advection/advection_mpi
-binary/binary_mpi
-binary_adi/binary_adi_mpi
```
pdmullen, line 18: If due to time constraints, do we think these are the best tests to eliminate for overall coverage? Or were these selections moreso made due to the time the test took?

pdmullen (approved):
Awesome, @brryan! 😸

tst/suites/parallel.suite
brryan: These are really nice tests but they are very slow on github runners. The GPU CI runs the serial versions of these, we may want to also have a darwin CPU CI to run expensive CPU MPI tests as well

tst/suites/gpu.suite
brryan: Yeah this takes like 2000 seconds on GPU, because its many cycles and few cells

tst/scripts/utils/artemis.py
brryan: Fixed

adamdempsey90:
We should probably also fix the last sections of the README

brryan:
> We should probably also fix the last sections of the README

Thanks for the reminder, fixed.

PR #13, opened 2024-11-20
Update to Parthenon with Kokkos 4.4.1

brryan:
## Background

Chicoma and Venado are failing at runtime when running with multiple GPUs using CUDA-aware MPI. Forrest found that moving to Kokkos 4.4.1 fixes this issue at least on Venado. 

## Description of Changes

- [x] Switch our Parthenon submodule to the current branch that includes Kokkos 4.4.1 support.
- [x] Update `env/bash` script to support Venado (sort of)
- [x] Fix some compiler error (`constexpr if` capture)and warning (unused var) on recent `nvcc` 

## Checklist

- [x] [Parthenon PR](https://github.com/parthenon-hpc-lab/parthenon/pull/1191) is merged


pdmullen (approved):
LGTM!

env/bash
```diff
@@ -27,9 +27,9 @@ if [[ $HOSTNAME == ch-fe*  ||  $SLURM_CLUSTER_NAME == "chicoma" ]]; then
     else
         PARTITION="chicoma-cpu"
     fi
-elif [[ "$HOSTNAME" =~ ^ve-rfe[1-3]$ || ( $SLURM_CLUSTER_NAME == "venado" && $SLURM_JOB_PARTITION == "gpu" ) ]]; then
+elif [[ "$HOSTNAME" =~ ^ve-rfe[1-3]$ || "$HOSTNAME" =~ ^ve-fe[1-3]$ || ( $SLURM_CLUSTER_NAME == "venado" && $SLURM_JOB_PARTITION == "gpu" ) ]]; then
     PARTITION="venado-gh"
-elif [[ "$HOSTNAME" =~ ^ve-rfe[4-7]$ || ( $SLURM_CLUSTER_NAME == "venado" && $SLURM_JOB_PARTITION == "cpu" ) ]]; then
+elif [[ "$HOSTNAME" =~ ^ve-rfe[4-7]$ || "$HOSTNAME" =~ ^ve-fe[4-7]$ || ( $SLURM_CLUSTER_NAME == "venado" && $SLURM_JOB_PARTITION == "cpu" ) ]]; then
```
adamdempsey90: I don't think you can distinguish cpu/gpu from the hostname. You can be on a grace-grace frontend but submit to a grace-hopper backend. The hostname you check on the backend is still the grace-grace one. I think if you did hostname in [1-3] or  `$SLURM_GPUS_ON_NODE > 0`, that would always work

adamdempsey90 (approved):

env/bash
brryan: Yeah you're right, thanks for noticing this. I thought at some point in the past `HOSTNAME` wasn't defined on venado backends? But it definitely is now, maybe I'm just misremembering. Yes I can update this logic to fix this.

This also doesn't work for e.g. `SLURM_JOB_PARTITION=gpu_debug` (it's only been sneaking through because I always use gpu frontends for gpu backends etc.)

brryan: OK this should be fixed now. I tested it on cpu frontend, and gpu backend via either cpu frontend or gpu frontend. I'm not sure I have access to the CPU backends actually to test those

PR #14, opened 2024-11-21
Derived test module reloads and cleanup

pdmullen:
## Background

Moderate cleanup of CI and fixes regression execution on Macs.

## Description of Changes

- Reload "base" test module in derived tests via `importlib` (such that we will not worry about future derived tests having shared states dependent on order of execution).
- Fix Mac regression execution, requires special care of positioning of `librebound.so`.  
- Remove duplicated code in tests.
- Fix a few edge cases in CI workflows (hopefully without breaking others...)

## Checklist

### DIFF
--- a/tst/run_tests.py
+++ b/tst/run_tests.py
@@ -48,16 +48,9 @@ import scripts.utils.artemis as artemis  # noqa
 logger = logging.getLogger("artemis")
 
 
-def read_cmakecache(filename):
-    with open(filename, "r") as f:
-        for line in f.readlines():
-            if "artemis_BINARY_DIR" in line:
-                return os.path.join(line.split("=")[-1].strip(), "src", "artemis")
-
-
 def process_suite(filename):
     tests = []
-    fname = os.path.join(artemis.get_source_directory(), "tst", "suites", filename)
+    fname = os.path.join(artemis.get_artemis_dir(), "tst", "suites", filename)
 
     with open(fname, "r") as f:
         for line in f.readlines():
@@ -78,7 +71,7 @@ def process_suite(filename):
                                 # path=["scripts/" + test_name], prefix=test_name + "."
                                 path=[
                                     os.path.join(
-                                        artemis.get_source_directory(),
+                                        artemis.get_artemis_dir(),
                                         "tst",
                                         "scripts",
                                         test_name,
@@ -104,7 +97,7 @@ def main(**kwargs):
                     for _, name, _ in iter_modules(
                         path=[
                             os.path.join(
-                                artemis.get_source_directory(),
+                                artemis.get_artemis_dir(),
                                 "tst",
                                 "scripts",
                                 directory,
@@ -128,7 +121,7 @@ def main(**kwargs):
                     for _, name, _ in iter_modules(
                         path=[
                             os.path.join(
-                                artemis.get_source_directory(), "tst", "scripts", test
+                                artemis.get_artemis_dir(), "tst", "scripts", test
                             )
                         ],
                         prefix=test + ".",
@@ -164,10 +157,10 @@ def main(**kwargs):
                 logger.warning("Unable to " 'import "{:}".'.format(missing_module))
                 deps_installed = False
         if not deps_installed:
-            logger.warning("WARNING! Not all required Python modules " "are available")
+            logger.warning("WARNING! Not all required Python modules are available")
 
         # Build Artemis
-        if not artemis.custom_exe and not kwargs.pop("reuse_build"):
+        if not artemis.get_supplied_exe() and not kwargs.pop("reuse_build"):
             try:
                 os.system("rm -rf {0}/build".format(current_dir))
                 # insert arguments for artemis.make()
@@ -215,7 +208,7 @@ def main(**kwargs):
             # For CI, print after every individual test has finished
             logger.info("{} test: run(), analyze() finished".format(name))
     finally:
-        if not kwargs.pop("save_build") and not artemis.custom_exe:
+        if not kwargs.pop("save_build") and not artemis.get_supplied_exe():
             os.system("rm -rf {0}/build".format(current_dir))
 
     # Report test results
@@ -269,7 +262,7 @@ def log_init(args):
     # setup log_file
     log_fn = kwargs.pop("log_file")
     if log_fn:
-        f_handler = logging.FileHandler(os.path.join(artemis.artemis_log_dir, log_fn))
+        f_handler = logging.FileHandler(os.path.join(artemis.get_log_dir(), log_fn))
         f_handler.setLevel(0)  # log everything
         f_format = logging.Formatter(
             "%(asctime)s|%(levelname)s" ":%(name)s: %(message)s"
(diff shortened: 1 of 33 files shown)


tst/scripts/utils/artemis.py
```diff
@@ -96,18 +119,19 @@ def make(cmake_args, make_nproc):
...
-    input_filename_full = os.path.join(artemis_inputs_dir, input_filename)
-    run_command += ["-i", input_filename_full]
+        run_command += ["-r", os.path.join(artemis_data_dir, restart)]
+    run_command += ["-i", os.path.join(artemis_inputs_dir, input_filename)]
+    run_command += ["-d", artemis_data_dir]
 
     try:
-        # os.chdir(run_directory)
-        os.chdir(artemis_run_dir)
+        os.chdir(artemis_exe_dir)
```
adamdempsey90: I would prefer to keep the run directory as it was and instead copy `librebound.so` from the exe directory to the run directory. I assume things like core dumps would get placed in the exe directory in this model?

I would also prefer to not use `-d` and have the snapshots in the run directory.

tst/scripts/utils/artemis.py
```diff
@@ -122,6 +146,131 @@ def run(nproc, input_filename, arguments, restart=None):
...
+    )
+    cb.ax.xaxis.set_ticks_position(where)
+    cb.ax.xaxis.set_label_position(where)
+    cb.ax.tick_params(labelsize=labelsize)
+
+    return cb
+
+
+# Shared function to compute analytic Ogilvie & Lubow (2002) spiral positions
+def spiral_pos(r, r0=1.0, p0=np.pi, h=0.05):
```
adamdempsey90: This doesn't belong in `artemis.py`. It belongs in `binary.py`, and other files can import `binary` to use it if they want it.

tst/scripts/utils/artemis.py
```diff
@@ -18,48 +18,71 @@
...
+artemis_data_dir = "/dev/null"
+artemis_fig_dir = "/dev/null"
+artemis_log_dir = "/dev/null"
+use_mpi_oversubscribe = False
+use_supplied_exe = False
+
+
+# Optionally invoke MPI oversubscribe
+def set_mpi_oversubscribe(use_oversubscribe):
+    global use_mpi_oversubscribe
```
adamdempsey90, line 41: IMO dealing with these global variables isn't great. It would be great if we simplified this part with just using normal inheritance. But I understand that's a significant rewrite.

adamdempsey90 (changes requested):
Maybe some minor changes

tst/scripts/utils/artemis.py
pdmullen: I believe the current design does everything the earlier iteration did (plus some edge case fixes) without a manual copy of `librebound.so`.  

I like that this design gives `parthenon` the control of routing outputs using its `-d` flag.  

Also, I believe the core dump files belong wherever the executable is.  

You can routed the snapshots wherever you like with `--output_dir` in the current design.

I have also printed every path possible (there are too many...) so that you can clearly see where everything goes.

tst/scripts/utils/artemis.py
pdmullen: One test should not load another test (with the exception of derived tests).  

Moving this to a shared place prevents code duplication. 

I am open to moving this somewhere besides `artemis.py`, but not having e.g., `disk_nbody` load `binary`.

tst/scripts/utils/artemis.py
pdmullen: I didn't introduce this pattern --- hopefully I cleaned it up a bit.

tst/scripts/utils/artemis.py
adamdempsey90: Sure, but at the expense of running in the executable directory (and needing to use `-d`), which I was trying hard to avoid doing in the last iteration of this. The reason to not do that is if someone wants to run the tests on a released executable. They should not be running inside that shared directory -- especially if there are going to be things written to that directory.

tst/scripts/utils/artemis.py
adamdempsey90: I wouldn't call 
```
from binary import spiral_pos
```
loading the test, but loading an analysis function

tst/scripts/utils/artemis.py
pdmullen: Even `main` defaults to outputting to `exe_path`:

https://github.com/lanl/artemis/blob/develop/tst/run_tests.py#L307

I am open to changing this to `cwd`

tst/scripts/utils/artemis.py
adamdempsey90: This was a general comment on using these global variables and wasn't aimed at this changeset.

brryan: Yeah this was my mistake introducing this before. I think if we make this file into an `artemis` class rather than using it as a module that would fix this, but that's a fair amount of typing and probably best left for another PR

tst/scripts/utils/artemis.py
brryan: Good point about `main` outputting to `exe_path`, I think that is a mistake and I agree with Adam that 1) librebound.so goes to the run directory 2) we run from the run directory (so core dumps etc. go there) 

And we can still use `-d` and `--output_dir` separately if necessary right? Maybe I'm missing a subtlety there

adamdempsey90: that was a mistake, then.

tst/scripts/utils/artemis.py
```diff
@@ -122,6 +148,130 @@ def run(nproc, input_filename, arguments, restart=None):
         os.chdir(current_dir)
 
 
+# Function for reading Artemis data
+def load_level(n, base, dir="./", off=0):
```
brryan: Should we just import `ahdf.py` to do this work? 

I already use `ahistory.py` elsewhere in the tests for analysis.

tst/scripts/utils/artemis.py
brryan: `tst/scripts/utils/analysis.py`?

tst/scripts/utils/artemis.py
adamdempsey90: Maybe, this was me just copying my normal loader for uniform / nested refinement meshes. I don't know what all goes on in `ahdf.py` since I don't use it.

tst/scripts/utils/artemis.py
adamdempsey90: Rather than putting test specific things in something like `utils/analysis.py`, I would prefer to restructure the tests a bit. For example, all tests of binaries go in a `tst/scripts/binary/` directory and they all can share an `../binary/analysis.py` file with binary-specific functions. Or all tests using a specific pgen go in `tst/scripts/pgen` (I slightly prefer this).

tst/scripts/utils/artemis.py
brryan: OK I see now that this was moved, not added, thanks. 

I do think that within the artemis repository we should have a single maintained path for interpreting hdf5 outputs, and that is `ahdf.py` (which is derived from `phdf.py` and in that sense is "parthenonic"). But that is a change for another PR.

tst/scripts/utils/artemis.py
brryan: I think a general hierarchical model like
```
tst/scripts/
  problemA/
    problemA_var1.py
    problemB_var2.py
    problemA_analysis.py
  problemB/
    problemB.py
  utils/
    analysis.py // routines used by both problemA and problemB
```
is fine, I don't have a strong preference for names either exactly matching or not exactly matching pgen names, but I do think shared utility functions should 1) live at a directory level common to all tests that use them 2) have unique names between different analysis scripts so we dont run into issues with `import analysis` all over the place confusing python

pdmullen: > tst/scripts/utils/analysis.py

Open to this.  

> Rather than putting test specific things in something like utils/analysis.py, I would prefer to restructure the tests a bit. For example, all tests of binaries go in a tst/scripts/binary/ directory and they all can share an ../binary/analysis.py file with binary-specific functions. Or all tests using a specific pgen go in tst/scripts/pgen (I slightly prefer this).

Or this.

tst/scripts/utils/artemis.py
brryan: Deprecated by previous discussion

tst/scripts/utils/artemis.py
brryan: Especially if moving to a hierarchy of analysis routines, should we even have a `utils` dir? `utils/analysis.py` and `utils/artemis.py` could just live in `tst/scripts`?

pdmullen: For this MR, can we stick with moving these shared functions to a shared `tst/scripts/utils/analysis.py`? I am not opposed to the hierarchical approach, but this (1) will change how we glob for tests (we will need to be careful here...) and (2) will change how we import.  It is probably worth thinking through where the analysis scripts live too, as I like the current design where every `.py` file in the `scripts` subdirectory is a test (well... with the exception of `utils`).  Maybe something like `pgen_name/analysis/analysis.py`? This might make the `glob` easier too.  

I already sunk more time into this PR than you probably wanted me to 😅 And even though the analysis treatment is not finalized, it did eliminate significant code duplication.

tst/run_tests.py
```diff
@@ -279,61 +263,85 @@ def log_init(args):
...
-        if output_dir is None:
-            output_dir = os.getcwd()
         # If we are in a directory with an executable, default to using that
-        local_path = os.path.join(os.getcwd(), "artemis")
-        if os.path.exists(local_path) and os.access(local_path, os.X_OK):
-            print(f"Found local executable {local_path}")
-            print(f"Outputting results to {output_dir}")
-            artemis.set_executable(local_path, output_dir)
+        local_path = os.path.join(cwd, "artemis")
+        if os.path.isfile(local_path) and os.access(local_path, os.X_OK):
```
brryan: Per previous discussion I do think we should still guard this auto-find logic behind an optional `--cwd` flag, interpreted as
```python
    parser.add_argument(
        "--cwd",
        action="store_true",
        help="search local path for preexisting executable",
    )
```
And then the logic here should I think be
```python
if exe_path is not None:
  # same
else:
  use_cwd = kwargs.pop("cwd")
  if use_cwd:
    if os.path.isfile(local_path) and os.access(local_path, os.X_OK):
      # same, run code in `cwd()/testing`
    else:
      # Same, check for `CMakeCache.txt` in `cwd()`, fail if not found
  else:
    # Same, build code and run in `artemis/tst/build/src/artemis`
```

This avoids all the special cases we were discussing before about trying to run the code with default behavior and having it unintentionally pick up on a nearby `artemis` executable.

brryan (approved):
Apart from the `--cwd` option to guard against surprises with searching for local executables which I would like to see added, LGTM, approving now, thanks for all this typing @pdmullen, sorry I broke your workflow with `--oversubscribe` before.

tst/scripts/utils/artemis.py
brryan: Yeah sorry I should've added here, I don't think this is a change for this PR.

pdmullen:
> Apart from the `--cwd` option to guard against surprises with searching for local executables which I would like to see added, LGTM, approving now, thanks for all this typing @pdmullen, sorry I broke your workflow with `--oversubscribe` before.

Addressed in [5e90a73](https://github.com/lanl/artemis/pull/14/commits/5e90a734465a0f1a5302ad1fdb84baa7912afb1a)

PR #22, opened 2024-12-02
Preserve CI artifacts 

### LINKED ISSUE #16: Have darwin CI store artifacts locally
The Darwin CI script reports the result of the test run to github, but it does not store any artifacts from the run, which would be useful for diagnosing failures. 

I don't know how to send artifacts from Darwin to github. But one possibility would be to have each CI run create a `ci_darwin_volta-x86_PR[#]_[slurm job ID]` directory in `artemis/tst` (by default, optionally overwritten) that includes the slurm output file already created, and also all the artifacts created by the testing. 

I will get to this after:
- [ ] PR #14 is merged

brryan:
## Background

Right now, because the Darwin CI runs in a temporary directory, all artifacts are lost. 

This PR sends artifacts to `/usr/projects/jovian` so all LANL people with Darwin access and `jovian` group membership can see them. 

Fixes #16 

## Description of Changes

- [x] Create a `/usr/projects/jovian/ci/[moniker]/pr[pr number]` directory and use that for output in Darwin CI script
- [x] Update permissions of artifacts for `jovian` group access
- [x] Move SLURM output to artifacts directory

## Checklist


pdmullen (approved):

tst/launch_ci_runner.py
```diff
@@ -90,10 +90,15 @@ def run_tests_in_temp_dir(pr_number, head_repo, head_ref, commit_sha):
...
-                + " --log_file=ci_cpu_log.txt",
+                + f" --output_dir={output_dir}"
+                + " --log_file=darwin_log.txt",
             ]
             ret = subprocess.run(test_command, check=True)
 
+            # Set permissions to output files
+            subprocess.run(["chgrp", "-R", "jovian", output_dir], check=True)
+            subprocess.run(["chmod", "-R", "750", output_dir], check=True)
+
```
adamdempsey90: If the test fails, then the permissions never get changed?

adamdempsey90: maybe set the permissions when you create the directory, rather than after things run? https://superuser.com/questions/808345/change-default-permissions-for-new-files-and-folders

brryan: Good catch! You're right. I think it's fine to just remove the exception and check the error code manually (as done now)? I don't think `subprocess.run` can actually crash the program to the point that we don't end up updating the github CI later on

tst/launch_ci_runner.py
```diff
@@ -75,28 +75,33 @@ def run_tests_in_temp_dir(pr_number, head_repo, head_ref, commit_sha):
...
+        subprocess.run(["chgrp", "-R", "jovian", output_dir], check=True)
+        subprocess.run(["chmod", "-R", "750", output_dir], check=True)
+
+        if ret.returncode == 0:
             # CI apparently succeeded; indicate that
             return True
-        except subprocess.CalledProcessError:
+        else:
             # If CI failed, indicate that
             return False
```
adamdempsey90: Just return the returncode? `return ret.returncode == 0`

tst/launch_ci_runner.py
```diff
@@ -177,15 +184,28 @@ def run_tests_in_temp_dir(pr_number, head_repo, head_ref, commit_sha):
...
 
+            # Build output path and create directory if necessary
+            username = os.getenv("USER")
+            output_dir = os.path.join(
+                "/usr",
+                "projects",
+                "jovian",
+                "ci",
+                username,
+                f"pr_{args.pr_number}",
```
adamdempsey90: Is there any reason to have the username in the directory path? Since everything is open permissions to the group, I'm not seeing a good reason why all the CI jobs for a given PR shouldn't just be directly in `/usr/projects/jovian/ci/pr_X`

brryan: There are a couple competing issues here:
- Multiple users running the CI script on the same could overwrite each others' artifacts (potentially with different commits)
- We don't want to leave every CI run's artifacts in `/usr/projects/jovian` forever
- We want anyone to be able to quickly find the latest CI output regardless of user (an issue with my current approach)

What about writing outputs to the directory `/usr/projects/jovian/ci/pr_[pr number]/[date]_[time]` and pairing that with a `/usr/projects/jovian/ci/cleanup.py` script anyone can call that will erase all `[date]_[time]` folders older than seven days? The latter script could be linked to a cron job except I'm not sure that we users can create cron jobs on darwin so it might have to be run manually every so often.

adamdempsey90: yes, `/usr/projects/jovian/ci/pr_[pr number]/[date]_[time]` is what I was thinking of.

brryan: Fixed:
```
[brryan@darwin-fe1 2024-12-02_12:54:58]$ pwd
/usr/projects/jovian/ci/pr_22/2024-12-02_12:54:58
[brryan@darwin-fe1 2024-12-02_12:54:58]$ ls
artemis_ci_darwin_volta-x86_PR22_16115048.out
```

adamdempsey90 (approved):

brryan:
Just one more issue I'll fix before merge -- right now we save all the `data` output files by default, which ends up being `734 MB` per CI run in `/usr/projects/jovian` which we can't afford. I'll add an option to remove the data files after they're created.

tst/run_tests.py
```diff
@@ -221,6 +221,11 @@ def main(**kwargs):
         "Summary: {0} out of {1} {2} "
         "passed\n".format(num_passed, num_tests, test_string)
     )
+
+    # Erase simulation output data unless requested that it be saved
+    if kwargs.pop("erase_data"):
+        os.system(f"rm -rf {artemis.artemis_data_dir}")
```
pdmullen: open to changing to `artemis.get_artemis_dir()` for consistency with elsewhere?

pdmullen (approved):

tst/run_tests.py
brryan: I wanted to leave stuff like figures and logs, only erase the `*.phdf` outputs from artemis that get written to `artemis_data_dir`. So I think this naming is OK as is?

pdmullen: Ah sorry... I am still waking up this morning... 

I meant using `artemis.get_data_dir()`

i.e., just using the getter function

brryan: Ah yes you're right, I made that switch to the getter function, thanks

PR #24, opened 2024-12-03
Unit system

brryan:
## Background

We need to be able to set up a common conversion factor between code and physical units

## Description of Changes

- [x] Add `ArtemisUtils;:Units` class that contains unit conversions and physical constants in the physical unit system, set up from parameter input
- [x] Multiple ways to initialize code units (length/time/mass, star mass + orbital radius)
- [x] Ability to modify physical constants manually 
- [x] Report unit conversion in params so dump files include this information for analysis script usage?

## Checklist


src/utils/units.cpp
```diff
@@ -0,0 +1,100 @@
...
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//========================================================================================
+
+#include "units.hpp"
+
+namespace ArtemisUtils {
+
+Units::Units(ParameterInput *pin) {
+  unit_system_ = pin->GetOrAddString("artemis/units", "type", "code");
```
adamdempsey90: I'd slightly prefer the input block be `<units>` and not `<artemis/units>` since we don't really have `artemis/*` blocks.

src/utils/units.cpp
```diff
@@ -0,0 +1,100 @@
...
+  if (unit_system_ == "code") {
+    length_ = 1.;
+    time_ = 1.;
+    mass_ = 1.;
+    G_ = 1.;
+    kb_ = 1.;
```
adamdempsey90, line 70: I don't know if code units is `kb=1` or `Rgas=1`. I tend to use `Rgas=1` (and I think the default `cv` assumes that right now).

src/utils/units.cpp
```diff
@@ -0,0 +1,100 @@
...
+    c_ = 1.;
+    h_ = 1.;
+    Msolar_ = 1.;
+    AU_ = 1.;
+  } else {
+    if (unit_system_ == "cgs") {
+      parthenon::constants::PhysicalConstants<parthenon::constants::CGS> pc;
```
adamdempsey90: Should we be grabbing all the constants in parthenon's constants?

src/utils/units.cpp
```diff
@@ -0,0 +1,100 @@
...
+  if (unit_system_ != "code") {
+
+    unit_specifier_ = pin->GetOrAddString("artemis/units", "specifier", "base");
+    if (unit_specifier_ == "base") {
+      length_ = pin->GetReal("artemis/units", "length");
+      time_ = pin->GetReal("artemis/units", "time");
+      mass_ = pin->GetReal("artemis/units", "mass");
+    } else if (unit_specifier_ == "orbit") {
+      const Real Mstar = pin->GetReal("artemis/units", "star_mass");     // Solar masses
+      const Real Rorbit = pin->GetReal("artemis/units", "orbit_radius"); // AU
```
adamdempsey90: I would call these total mass and reference radius

src/utils/units.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <cmath>
+
+namespace ArtemisUtils {
+
+class Units {
+ public:
+  Units(ParameterInput *pin);
+
+  // Unit conversions
+  Real GetLengthCodeToPhysical() const { return length_; }
```
adamdempsey90, line 41: Should these all be accessible on device?

src/nbody/nbody.cpp
```diff
@@ -84,7 +87,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   Real Rf[3] = {0.0};
   Real Vf[3] = {0.0};
   if (global_frame && (Omf != 0.0) && (qshear != 0.0)) {
-    const Real R0 = std::pow(SQR(Omf) / GM, 1.0 / 3.0);
+    const Real R0 = std::pow(SQR(Omf) / units.GetGCode(), 1.0 / 3.0);
```
adamdempsey90: This should really be GM not G

src/nbody/nbody.cpp
```diff
@@ -70,11 +72,12 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
   params.Add("dt_reb", dt_reb);
 
-  // Unit system for gravity
-  const Real GM = pin->GetReal("gravity", "gm");
-  params.Add("GM", GM);
+  // Scaling factor for gravitational forces
   params.Add("mscale", pin->GetOrAddReal("nbody", "mscale", 1.0));
 
+  // Reference gravitational mass in code units (where mass = 1)
+  const Real GM = units.GetGCode();
```
adamdempsey90: This should really be GM not G

src/nbody/nbody.cpp
```diff
@@ -70,11 +72,12 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   if (integrator == "none") dt_reb = Big<Real>();
   params.Add("dt_reb", dt_reb);
 
-  // Unit system for gravity
-  const Real GM = pin->GetReal("gravity", "gm");
```
adamdempsey90, line 74: Other packages like viscosity and maybe drag grab GM from here. So they'll need to be changed as well.

adamdempsey90 (commented):
How are we going to access the units on device?

src/utils/units.cpp
```diff
@@ -0,0 +1,100 @@
...
+  c_code_ = c_ * time_ / length_;
+  h_code_ = h_ * time_ / mass_ * std::pow(length_, -2);
+  Msolar_code_ = Msolar_ / mass_;
+  AU_code_ = AU_ / length_;
+
+  if (unit_system_ != "code") {
+
+    unit_specifier_ = pin->GetOrAddString("artemis/units", "specifier", "base");
+    if (unit_specifier_ == "base") {
+      length_ = pin->GetReal("artemis/units", "length");
```
pdmullen: I think it would be really hard to handle all combinations... but it would be nice to also do something like specify a density unit and length unit.  Wonder if there is a clever way of doing something like that.

brryan: I think you would just have to find a descriptive name for that approach for describing units and then compare `unit_specifier_` with that and then look for the appropriate input parameter constants. No real limit on how many such specifiers we can have, but I don't immediately see anything more clever than that.

src/utils/units.cpp
brryan: My feeling was that `artemis/units` is the appropriate scope for these parameters, since 1) the units are used artemis-wide 2) other parthenon codes might have their own notion of `units` input parameters and this would avoid crosstalk between packages in different codes. 

We already have `artemis/` parameters so I don't see a style issue with having blocks as well there, but if you foresee issues with doing this lets keep discussing.

src/utils/units.cpp
pdmullen: I'd personally prefer never to think about $R$ 😬  My brain is too hardwired to:

rho k T / (mu mp) == gm1 rho e

src/utils/units.cpp
```diff
@@ -0,0 +1,100 @@
...
+  if (pin->DoesParameterExist("artemis/units/constants", "Msolar")) {
+    Msolar_ = pin->GetReal("artemis/units/constants", "Msolar");
+  }
+  if (pin->DoesParameterExist("artemis/units/constants", "AU")) {
+    AU_ = pin->GetReal("artemis/units/constants", "AU");
+  }
+
+  // Convert constants to code units
+  G_code_ = G_ * std::pow(length_, -3) / mass_ * std::pow(time_, 2);
+  kb_code_ = std::pow(time_, 2) / mass_ * std::pow(length_, -2); // 1 K = 1 code unit temp
```
brryan:
```suggestion
  kb_code_ = kb_ * std::pow(time_, 2) / mass_ * std::pow(length_, -2); // 1 K = 1 code unit temp
```

adamdempsey90:
Two design comments. 

(1) It almost feels like your "units" struct should really be called `PhysicalConstants` or something like that. The units part could maybe be separated into functions that do the conversions, e.g., `ConvertOpacityToCGS`. 

(2) I don't know if we need a full `<artemis/units>` node. Following the coordinate specification maybe just something in `<artemis>` directly,

```
<artemis>
units = cgs  # cgs, ppd, user
length = ...
mass   = ...
time    = ...
```
I also don't think we necessarily need the ability to specify every physical constant in the input deck. I would like to discourage that if at all possible.

src/utils/units.cpp
brryan: Default units here is `kb = mp = 1`; I think `kb` makes more sense in this case because singularity-eos interacts with the specific heat capacity `<c_v> = erg / (g * K) -> c_v = k_b / (gam - 1) mu m_p`.

You are right that I need to actually calculate `c_v` correctly as part of this PR

src/utils/units.cpp
brryan: I was just taking what we need; parthenon's constants is pretty thorough. Also because this class is runtime and be copied to device, it probably is best to keep it as small as possible.

src/utils/units.cpp
brryan: Yes good point, in particular `total_mass` is actually correct based on how `nbody` already works. Fixed.

src/utils/units.hpp
brryan: Yes; done. Class can only be constructed on device though because it interacts with `ParameterInput`. Device-usable copy constructor provided for getting this into kernels though.

brryan:
> Two design comments.
> 
> (1) It almost feels like your "units" struct should really be called `PhysicalConstants` or something like that. The units part could maybe be separated into functions that do the conversions, e.g., `ConvertOpacityToCGS`.
> 
> (2) I don't know if we need a full `<artemis/units>` node. Following the coordinate specification maybe just something in `<artemis>` directly,
> 
> ```
> <artemis>
> units = cgs  # cgs, ppd, user
> length = ...
> mass   = ...
> time    = ...
> ```
> 
> I also don't think we necessarily need the ability to specify every physical constant in the input deck. I would like to discourage that if at all possible.

1) Yeah I think really this should be two different classes (now done). Only slight issue is a circular dependency with solar mass/AU in setting the unit system via reference mass/radius that I had to dance around. 

2) I'll defer to you on whether `units` is a node; it's gone now.

3) Yeah I removed the ability to customize physical constants. This was always intended as a "for experts only" option but it doesn't even work right now with our made up constants for e.g. the rad shocktubes because that's constexpr so I'm deferring the custom constants issue now.

src/nbody/nbody.cpp
brryan: The `M` here is the sum of the particles' `p.m` values right? So should this value be set in `NBodySetup` as `units.GetGCode() * mtot`?

adamdempsey90: Since `GM` is really set at startup and put into params, I don't think we need any units stuff in places where it's used. When it's initially defined it should probably just be `GM = units.GetGCode() * units.GetMCode() * GM`.

src/utils/units.hpp
```diff
@@ -0,0 +1,142 @@
...
+  Real GetAUCode() const { return AU_code_; }
+
+ private:
+  // Physical constants in physical units
+  Real G_;      // Gravitational constant
+  Real kb_;     // Boltzmann constant
+  Real c_;      // Speed of light
+  Real h_;      // Planck constant
+  Real Msolar_; // Solar mass
+  Real AU_;     // Astronomical unit
```
adamdempsey90: Other constants I'd like to see are:

- [ ]  `amu` and/or `mp` (cv = kb/(mu * amu *(g-1)))
- [ ]   `ev`
- [ ]  `year` 

At some point we'll also want  electron mass, but not at the moment. 

There's also `pc` if we want to be supportive AGN disks

src/gas/gas.cpp
```diff
@@ -96,7 +101,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   const std::string eos_name = pin->GetOrAddString("gas", "eos", "ideal");
   if (eos_name == "ideal") {
     const Real gamma = pin->GetOrAddReal("gas", "gamma", 1.66666666667);
-    const Real cv = pin->GetOrAddReal("gas", "cv", 1. / (gamma - 1.));
+    const Real mmw = pin->GetOrAddReal("gas", "mmw", 1.);
+    const Real cv = constants.GetKBCode() / ((gamma - 1.) * mmw);
```
adamdempsey90: This should be `mmw * amu`?

Also should we allow the user to specify `cv` separately from `mmw`, but default `cv` to this value?

src/utils/units.cpp
```diff
@@ -0,0 +1,98 @@
...
+                        std::pow(length_, 3)); // Orbital period
+    } else {
+      PARTHENON_FAIL("Unit specifier not recognized!");
+    }
+  }
+
+  // Remaining conversion factors
+  energy_ = std::pow(length_, 2) * mass_ * std::pow(time_, -2);
+  number_density_ = std::pow(length_, -3);
+  mass_density_ = mass_ * number_density_;
```
adamdempsey90: Others that may be useful are:
- [x] opacity
- [x] energy density and/or pressure
- [x] velocity
- [x] whatever units `cv` is

src/utils/units.hpp
```diff
@@ -0,0 +1,142 @@
...
+  UnitSystem GetUnitSystem() const { return unit_system_; }
+
+  // Unit conversions
+  KOKKOS_INLINE_FUNCTION
+  Real GetLengthCodeToPhysical() const { return length_; }
+  KOKKOS_INLINE_FUNCTION
+  Real GetLengthPhysicalToCode() const { return 1. / length_; }
+
+  KOKKOS_INLINE_FUNCTION
+  Real GetTimeCodeToPhysical() const { return time_; }
```
adamdempsey90, line 46: What do you think about these functions taking in the quantity and returning the converted quantity, as opposed to returning conversion factors and the caller applying them.

src/utils/units.hpp
```diff
@@ -0,0 +1,142 @@
...
+#ifndef UTILS_UNITS_HPP_
+#define UTILS_UNITS_HPP_
+
+#include "artemis.hpp"
+
+#include <cmath>
+
+namespace ArtemisUtils {
+
+enum class UnitSystem { scalefree, cgs };
```
adamdempsey90: A unit system for PPDs would be `Msolar` `AU` `yr/(2*pi)` and `K`. I think the CGS temperature unit makes it not "scalefree"? I would even call this the `ppd` unit system.

src/utils/units.cpp
```diff
@@ -0,0 +1,98 @@
...
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//========================================================================================
+
+#include "units.hpp"
+
+namespace ArtemisUtils {
+
+constexpr Real Msolar = 1.988416e33;
+constexpr Real AU = 1.495978707e13;
```
adamdempsey90, line 19: For Jupiter focused sims, would `Rjup` and `Mjup` also be useful to have?

src/utils/units.hpp
brryan: Added `amu`, `ev`, `year`, `pc`, good suggestions :+1:

src/gas/gas.cpp
brryan: Yeah better to specify this in units of `amu` than `g`. 

We might want to *also* allow the user to specify non-ideal cvs, specifically the Marshak EOS, so maybe we do want a `model` parameter here, e.g. `model = mmw, cv, marshak`. I'll try that.

adamdempsey90: I'd suggest just doing, 

```
const Real mmw = pin->GetOrAddReal("gas", "mmw", 1.);
const Real cv = pin->GetOrAddReal("gas", "cv", constants.GetKBCode() / (mmw * constants.GetAMUCode() * (gamma - 1.)));
```

We could back out the `mmw` if the user specified just `cv`, but I think in cases like that `mmw` is probably meaningless. 

Also, can we call `mmw` either `mu` (which is what it's called in the astro literature) or `abar` (which is what it's called in the lab codes)?

src/utils/units.hpp
brryan: I think this is a design choice that doesn't warrant a rewrite at this point. The one issue with the returning-converted-quantities approach is that you *need* functions for every kind of quantity you want to convert, unless you want to break the design if necessary by doing like
```c++
double l_unit = ConvertLengthCodeToCGS(1.);
double t_unit = ConvertTimeCodeToCGS(1.);
// Use some combination of powers of l_unit and t_unit to convert your custom quantity
```

src/utils/units.hpp
brryan: I'm not sure we need unit systems beyond scalefree and cgs; these really only serve to have scalefree problems with trivial physical constants, or have a conversion to physical units. Whether we convert to length in cm or length in m or length in AU is already handled by the conversion between code and physical units,.

src/utils/units.cpp
brryan: Agreed -- adding now

src/gas/gas.cpp
brryan: Switched `mmw` to `mu`, added optional `cv` override. I added logic so the user cant supply both `mu` and `cv` to avoid inconsistencies.

I was wrong about also supporting the Marshak EOS here, that will have to be a separate singularity-eos variant.

src/utils/units.cpp
brryan: Done :+1:

src/utils/units.hpp
adamdempsey90: I agree my choice comes with needing functions for every kind of quantity. I think that comes with safety, though, since there is one implementation of a unit conversion, and not multiple implementations. If a developer needs a conversion function that doesn't exist yet, they can add it.

src/utils/units.hpp
adamdempsey90: I suppose it's just a convenience in the input file if `units = ppd` populates all the `length`, `time`, `mass`, etc. for me.

src/utils/units.hpp
brryan: As discussed today let's keep the current function signatures.

src/utils/units.hpp
brryan: I added a `ppd` option for `<artemis> unit_system = ` with your proposed unit conversions so this convenience should actually be available that way. :+1:

src/nbody/nbody.cpp
```diff
@@ -70,9 +72,10 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
-  params.Add("GM", GM);
+  // Total mass of gravitating particles
+  const Real gm = constants.GetGCode() * pin->GetReal("gravity", "mass_tot") *
```
brryan: @adamdempsey90 OK I'm still thinking about this. I understand now how you provide nbody masses in terms of a portion of the total GM, but I think it's more consistent to specify total GM in terms of a total mass, and use the existing G in code units from our constants for consistency. I chose solar masses as the current unit for the total mass here; is that reasonable or would you prefer grams? Let me know if I'm still not understanding how this `GM` works but I think at this point this is consistent with the existing need for other packages to have a GM, and also consistent with our constants and units in this PR.

src/nbody/nbody.cpp
brryan: `gm` is again a `param`

src/nbody/nbody.cpp
adamdempsey90: I would think you would just multiply by `GetMCode`, not `GetMsolarCode`. `G` and `M` are set in such a way that in code units `GM=1` (for the `ppd` system).

adamdempsey90: Or maybe user inputs `GM` in code units and the mass unit, and then `G` is set from those two.

brryan: `G` is just `G`, so I strongly prefer respecting the value of that constant provided by the underlying scale (either `G=1` for `scalefree` or `G=6.67ewhatever` for `cgs` rather than letting the user override this in a maybe unexpected way (same way this PR no longer allows users to provide custom physical constants). With that requirement the only path forward I see is to have the user provide the total gravitating mass here, and let Artemis add the `G`. 

We can allow any units for `mass_tot` here, `Msolar` seemed like a reasonable choice since that's more likely to lead to round numbers in input files, but if you think there's a better default I don't have a strong opinion, except that I think we don't want the default to be in code units because code units aren't really for the user to interact directly with in general; we don't want people to have to juggle length/mass/time conversions when they're writing input failes. Also, for the `scalefree` backend `constants.GetMSolar() = 1` so in that case the user is just inputting trivially in code units (like current input decks). There isn't a `constants.GetMCode()` method because that would just return `1` (really it would be `units.GetMassCodeToCode()` which we dont provide because it's `=1`).

src/nbody/nbody.cpp
brryan: discussion moved below, resolving this

src/nbody/nbody.cpp
brryan: As discussed this afternoon in person, let's remove the `orbit` units with custom radius/gravitating mass, keep `ppd` units, and keep the `M` input parameter for specifiying the `GM` normalization since this mass is independent of the unit systems and must be specified somewhere in `nbody` gravity.

brryan:
@adamdempsey90 @pdmullen IMO this is ready to go; could I please get approvals, or more queries if there are outstanding issues? Thanks

src/gas/gas.cpp
```diff
@@ -136,10 +154,10 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   std::string scattering_model_name =
       pin->GetOrAddString("gas/opacity/scattering", "scattering_model", "none");
   if (scattering_model_name == "none") {
-    scattering = singularity::photons::GrayS(0.0, 1.0);
+    scattering = NonCGSUnitsS<GrayS>(GrayS(0.0, 1.0), time, mass, length, 1.);
   } else if (scattering_model_name == "constant") {
     const Real kappa_s = pin->GetOrAddReal("gas/opacity/scattering", "kappa_s", 0.0);
-    scattering = singularity::photons::GrayS(kappa_s, 1.0);
+    scattering = NonCGSUnitsS<GrayS>(GrayS(kappa_s, 1.0), time, mass, length, 1.);
```
adamdempsey90, line 160: Just to be clear, even if artemis is using cgs units, this is still called `NonCGSUnits` right?

src/nbody/nbody.cpp
```diff
@@ -84,7 +87,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   Real Rf[3] = {0.0};
   Real Vf[3] = {0.0};
   if (global_frame && (Omf != 0.0) && (qshear != 0.0)) {
-    const Real R0 = std::pow(SQR(Omf) / GM, 1.0 / 3.0);
+    const Real R0 = std::pow(SQR(Omf) / constants.GetGCode(), 1.0 / 3.0);
```
adamdempsey90: Why change this from `GM`? `GM` should already be in code units, so it was fine before right?

src/utils/units.cpp
```diff
@@ -0,0 +1,115 @@
...
+    std::string unit_specifier = pin->GetOrAddString("artemis", "specifier", "base");
+    if (unit_specifier == "base") {
+      length_ = pin->GetReal("artemis", "length");
+      time_ = pin->GetReal("artemis", "time");
+      mass_ = pin->GetReal("artemis", "mass");
+    } else if (unit_specifier == "ppd") {
+      length_ = AU;
+      mass_ = Msolar;
+      time_ = Year / (2. * M_PI);
+    } else {
```
adamdempsey90: I don't understand the difference between `artemis/specifier` and `artemis/units/type`? I thought it would just be 

```
<artemis>
...
units = scalefree  # scalefree, cgs, ppd
length = ..
mass = ...
time = ...
```

src/utils/units.cpp
```diff
@@ -0,0 +1,115 @@
...
+      PARTHENON_FAIL("Unit specifier not recognized!");
+    }
+  }
+
+  // Remaining conversion factors
+  energy_ = std::pow(length_, 2) * mass_ * std::pow(time_, -2);
+  number_density_ = std::pow(length_, -3);
```
adamdempsey90, line 58: Conversion factors for all the other compound quantities like `pressure`, `cv`, opacity, conductivity, viscosity?

src/artemis.cpp
```diff
@@ -52,6 +53,12 @@ Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
                         pin->GetInteger("parthenon/meshblock", "nx3")};
   artemis->AddParam("mb_dim", nb);
 
+  // Set up unit conversions for this problem
+  ArtemisUtils::Units units(pin.get(), artemis);
+  ArtemisUtils::Constants constants(units);
+  artemis->AddParam("units", units);
```
adamdempsey90, line 59: Why pass `artemis` to the `Units` constructor? You're storing the whole thing in params, so why store other members separately as well?

src/utils/units.cpp
adamdempsey90: I see below that you just compute those conversion on the fly. Not sure why some are computed on the fly and some are stored, but not a big deal

src/utils/units.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+  KOKKOS_INLINE_FUNCTION
+  Real GetCPhysical() const { return c_; }
+  KOKKOS_INLINE_FUNCTION
+  Real GetCCode() const { return c_code_; }
+
+  KOKKOS_INLINE_FUNCTION
+  Real GetHPhysical() const { return h_; }
+  KOKKOS_INLINE_FUNCTION
+  Real GetHCode() const { return h_code_; }
```
adamdempsey90, line 132: I don't know if every actually need `h`?

src/utils/units.cpp
```diff
@@ -0,0 +1,115 @@
...
+    c_ = pc.c;
+    h_ = pc.h;
+    amu_ = pc.amu;
+    eV_ = pc.eV;
+    Msolar_ = Msolar;
+    AU_ = AU;
+    Rjup_ = Rjup;
+    Mjup_ = Mjup;
+    pc_ = parsec;
+    Year_ = Year;
```
adamdempsey90, line 94: Sorry, one more constant that is being used is the radiation constant and optionally the steffan-boltzman constant.

src/nbody/nbody.cpp
brryan: Yes this was my leftover mistake, thanks for catching this, fixed.

src/gas/gas.cpp
brryan: Yes, this might be better called `CustomUnits` on the singularity side. I made the choice to always include the unit conversions to halve the number of singularity variants we're carrying around.

src/utils/units.cpp
brryan: `[scalefree, cgs]` and `[base, ppd]` are separate ideas; the first fixes the physical units through setting the values of fundamental constants (either all `1` or cgs values) and `[base, ppd]` is different ways of setting the conversion between physical and code units, with `base` requiring the user to set length, mass, and time conversions, and `ppd` is a default set of those conversions for convenience with no user-settable parameters.

But I didnt completely remove the `units` node here so I will turn this into `artemis/physical_units` and `artemis/unit_conversion`

src/utils/units.cpp
brryan: I tried to store the factors the need `std::pow` but calculate the others. It's just a choice I made.

src/artemis.cpp
brryan: It's crucial to store HDF5-outputtable conversion factors for use in analysis scripts, we could store those conversion factors separately here and not pass the artemis package but this felt like it kept `artemis.cpp` less messy.

src/utils/units.hpp
brryan: It may show up with energy <-> frequency for photons somewhere

src/utils/units.cpp
brryan: Good point, `ar` added

pdmullen (approved):
A huge effort. Thanks @brryan!

PR #26, opened 2024-12-05
Wrap rebound simulation pointer in struct

brryan:
## Background

We have a "memory leak" currently because we aren't properly freeing the rebound simulation struct when the `Params` object holding it expires. 

This may also help with the issue of having this parameter hold a different value on different processors? Needs to be tested. 

## Description of Changes

- [x] Wrap `reb_sim` in a struct with a custom destructor
- [x] Test MPI param consistency issue.
- [x] Merge [Parthenon PR](https://github.com/parthenon-hpc-lab/parthenon/pull/1215) that fixes MPI parameter consistency issue (actually unrelated to this rebound free issue)

## Checklist

### DIFF
--- a/src/nbody/nbody_utils.hpp
+++ b/src/nbody/nbody_utils.hpp
@@ -34,12 +34,30 @@ namespace NBody {
 extern void reb_extra_forces(struct reb_simulation *rsim);
 extern int collision_resolution(struct reb_simulation *const r, struct reb_collision c);
 
+// Container for reb_simulation pointer that allows it to be stored in Params with
+// automatic cleanup
+struct RebSim {
+
+  struct reb_simulation *&get() {
+    return reb_sim;
+  }
+
+  ~RebSim() {
+    if (reb_sim != nullptr) {
+      reb_simulation_free(reb_sim);
+    }
+  }
+
+ private:
+  struct reb_simulation *reb_sim = nullptr;
+};
+
 //----------------------------------------------------------------------------------------
 //! \fn  void NBody::SetReboundPtrs
 //! \brief
-static void SetReboundPtrs(struct reb_simulation *reb_sim) {
-  reb_sim->collision_resolve = collision_resolution;
-  if (RebAttrs::extras) reb_sim->additional_forces = reb_extra_forces;
+static void SetReboundPtrs(RebSim &reb_sim) {
+  reb_sim.get()->collision_resolve = collision_resolution;
+  if (RebAttrs::extras) reb_sim.get()->additional_forces = reb_extra_forces;
 }
 
 //----------------------------------------------------------------------------------------
@@ -75,7 +93,7 @@ static void PrintSystem(const int npart, ParArray1D<Particle> particles) {
 //----------------------------------------------------------------------------------------
 //! \fn  void NBody::SyncWithRebound
 //! \brief Copy the positions and velocities from the rebound sim to the Particles list
-static void SyncWithRebound(struct reb_simulation *r_sim, std::vector<int> particle_id,
+static void SyncWithRebound(RebSim r_sim, std::vector<int> particle_id,
                             ParArray1D<Particle> particles) {
   const auto npart = particle_id.size();
   auto particles_h = particles.GetHostMirrorAndCopy();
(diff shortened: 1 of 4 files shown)


src/nbody/nbody_utils.hpp
```diff
@@ -75,7 +93,7 @@ static void PrintSystem(const int npart, ParArray1D<Particle> particles) {
 //----------------------------------------------------------------------------------------
 //! \fn  void NBody::SyncWithRebound
 //! \brief Copy the positions and velocities from the rebound sim to the Particles list
-static void SyncWithRebound(struct reb_simulation *r_sim, std::vector<int> particle_id,
+static void SyncWithRebound(RebSim r_sim, std::vector<int> particle_id,
```
adamdempsey90:
```suggestion
static void SyncWithRebound(RebSim &r_sim, std::vector<int> particle_id,
```
Probably should always pass this by ref

src/nbody/nbody_utils.hpp
```diff
@@ -34,12 +34,30 @@ namespace NBody {
 extern void reb_extra_forces(struct reb_simulation *rsim);
 extern int collision_resolution(struct reb_simulation *const r, struct reb_collision c);
 
+// Container for reb_simulation pointer that allows it to be stored in Params with
+// automatic cleanup
+struct RebSim {
+
+  struct reb_simulation *&get() {
```
adamdempsey90: Can we add several constructors to this as well?
```
RebSim() {
 reb_sim = reb_simulation_create();
}
RebSim(struct rebound_simulation *r_sim) {
  reb_sim = reb_simulation_copy(r_sim);
}
RebSim(std::string filename) {
  reb_sim = reb_simulation_create_from_file(filename);
}
```

src/nbody/nbody_utils.hpp
adamdempsey90: and you might want to delete the copy constructors completely

brryan: I think it's fine to pass this by copy (maybe there was an issue with an older version, but now that I'm holding a `shared_ptr` that doesn't care about pass by copy vs reference (although reference is slightly more performant so I did switch). 

I think it's fine to keep the copy constructor now due to the `shared_ptr`.

src/nbody/nbody_utils.hpp
brryan: 1) Having the default constructor will do extra work of creating a rebound simulation that we discard for this case:
```c++
  RebSim reb_sim_rst;
  reb_sim_rst = nbody_pkg->Param<RebSim>("reb_sim");
```
so I don't think we want that unless I'm missing something
2) I added a custom copy constructor to avoid `.get()`s in the code
3) I added the std::string filename constructor
:+1:

adamdempsey90: Could you make a constructor that allows `RebSim reb_sim_rst(nbody_pkg->Param<RebSim>("reb_sim")`. 

The reason for 1 is just to have a true RAII class. Feels wrong to acquire the resource outside of the constructor.

src/nbody/nbody_utils.hpp
adamdempsey90: Okay, sounds good. I think you need to push those changes though, I don't see it.

src/nbody/nbody_utils.hpp
brryan: OK actually yeah I can just use 2) for the case you described. I did that and made the default constructor do a `reb_simulation_create`, I agree this is nicest.

src/nbody/nbody_utils.hpp
brryan: Pushed -- I was worried about
```c++
RebSim r_sim;
  if (parthenon::Globals::my_rank == 0) {
    // Advance the simulation.  If this is the final stage, advance the master simulation.
    // If not, advance a copy of the master.
    int sid = disable_stderr();
    r_sim = (stage < nstages) ? RebSim(reb_sim) : reb_sim;
```
but I think this is actually fine because for all nodes we later require an initialized `r_sim` when doing the MPI sync

src/nbody/nbody_utils.hpp
```diff
@@ -34,10 +36,48 @@ namespace NBody {
...
+  struct reb_simulation *get() const {
+    PARTHENON_REQUIRE(reb_sim, "Internal pointer is null!");
+    return reb_sim.get();
+  }
+
+  // Conversion operator to return the raw pointer
+  operator struct reb_simulation *() const { return reb_sim.get(); }
+
+  // Overloaded -> operator
+  reb_simulation *operator->() const { return reb_sim.get(); }
```
brryan, line 64: Some C++ trickery that lets us to `RebSim->integrator =` and `reb_simulation_function(RebSim, ...)` without the extra `.get()`s

src/nbody/nbody_utils.hpp
brryan: OK having a `reb_simulation_copy` inside the copy constructor appears to break the code during internal `Params` manipulations. I reverted to default copy constructor and added a `RebSim::copy` method, which is also a bit more explicit about what we're doing.

pdmullen (approved):

adamdempsey90 (approved):

PR #29, opened 2024-12-13
Dec 24 HPC fixes

brryan:
## Background

We've had runtime errors with GPU direct MPI on HPC machines for a little while. This has been resolved primarily to moving to Kokkos 4.5.

## Description of Changes

- [x] Update to latest Parthenon develop that includes Kokkos 4.5 as a submodule
- [x] Update to the latest Jaybenne version
- [x] Update Chicoma and Venado environments and build options to match currently working build/run pattern
- [x] Switch from `RelWithDebInfo` to `Release` to avoid slower compile times and large storage size of debug symbols for CUDA builds

## Checklist

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -102,11 +102,12 @@
         "CMAKE_MAKE_PROGRAM": "$env{MAKE_PROGRAM}",
         "CMAKE_BUILD_TYPE": "Debug",
         "Kokkos_ENABLE_DEBUG_BOUNDS_CHECK": "ON",
+        "CMAKE_C_COMPILER": "cc",
         "CMAKE_CXX_COMPILER": "$env{ARTEMIS_HOME}/external/parthenon/external/Kokkos/bin/nvcc_wrapper",
         "ARTEMIS_ENABLE_CUDA": "ON",
+        "Kokkos_ARCH_HOPPER90": "ON",
         "HDF5_INCLUDE_DIR": "$env{HDF5_ROOT}/include",
         "PARTHENON_DISABLE_HDF5_COMPRESSION": "ON",
-        "Kokkos_ARCH_HOPPER90": "ON",
         "NUM_GPU_DEVICES_PER_NODE": "4"
       }
     },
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -96,19 +102,21 @@
...
-        "Kokkos_ARCH_HOPPER90": "ON",
         "NUM_GPU_DEVICES_PER_NODE": "4"
       }
     },
     {
       "name": "venado-gpu-release",
       "cacheVariables": {
         "CMAKE_MAKE_PROGRAM": "$env{MAKE_PROGRAM}",
-        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_BUILD_TYPE": "Release",
```
adamdempsey90, line 118: Switching it to Release makes debuggers useless for finding problems. I don't know if that's worth the shorter compile times, though.

brryan: Yeah it's a tradeoff. My own thought is that if I hit a segfault I'm willing to rebuild in Debug mode to get symbols along with other stuff like bounds checking. But if you think always having the symbols in release-mode executables is advantageous, I'm fine with that and just let me know and I'll revert this change.

adamdempsey90: I suppose you can just recompile, yes. But if we're talking about a "release" Release, as in it's our official release and we don't change it, then that should 100% be RelWithDebInfo. But that can be changed by hand to include the debug symbols, so I think Release is fine here.

env/bash
```diff
@@ -121,37 +118,29 @@ elif [[ $PARTITION == "darwin-volta-x86" ]]; then
...
-    export MPICC=cc
-    export MPICXX=CC
-    export MPIFC=ftn
-    export CC=cc
-    export CXX=CC
-    export FC=ftn
-    export F90=ftn
+    unset OMP_NUM_THREADS
+    unset OMP_PROC_BIND
+    unset OMP_PLACES
```
adamdempsey90: Is something setting those? Our Kokkos doesn't include openmp does it?

adamdempsey90 (approved):

env/bash
brryan: I copied this over previously without checking that it's necessary; these indeed aren't set, so I removed the `unset` calls.

CMakePresets.json
brryan: OK good point I have on my todo list to make a release script which I agree makes sense to be `RelWithDebInfo` because compile times/executable sizes are not really a concern in that case whereas having debug symbols is.

PR #30, opened 2024-12-17
Permit pgen-defined custom source terms

pdmullen:
## Background
Permits custom pgen-defined source terms.  

## Description of Changes
Adds logic to enroll a custom pgen source term.  The task is called independent of enrollment, but checks for `nullptr`.

## Checklist

### DIFF
--- a/src/artemis_driver.cpp
+++ b/src/artemis_driver.cpp
@@ -234,10 +234,15 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
                                 time, bdt);
       }
 
+      // Apply pgen-defined source term
+      // NOTE(@pdmullen): Intentional positioning after S(W) srcs and before S(U) srcs
+      auto pgen_src =
+          tl.AddTask(rframe_src, ArtemisUtils::ProblemSourceTerm, u0.get(), time, bdt);
+
       // Apply drag source term
-      TaskID drag_src = rframe_src;
+      TaskID drag_src = pgen_src;
       if (do_drag) {
-        drag_src = tl.AddTask(rframe_src, Drag::DragSource<GEOM>, u0.get(), time, bdt);
+        drag_src = tl.AddTask(pgen_src, Drag::DragSource<GEOM>, u0.get(), time, bdt);
       }
 
       // Apply cooling source term
--- a/src/pgen/disk.hpp
+++ b/src/pgen/disk.hpp
@@ -815,10 +815,20 @@ void DiskBoundaryExtrap(std::shared_ptr<MeshBlockData<Real>> &mbd, bool coarse) 
 }
 
 //----------------------------------------------------------------------------------------
+//! \fn TaskStatus ProblemGeneratorSourceTerm()
+//! \brief Custom source term for disk pgen
+template <Coordinates GEOM>
+TaskStatus ProblemGeneratorSourceTerm(MeshData<Real> *md, const Real time,
+                                      const Real dt) {
+  PARTHENON_FAIL("Disk pgen-defined source term not yet implemented!");
+  return TaskStatus::complete;
+}
+
+//----------------------------------------------------------------------------------------
 //! \fn AmrTag ProblemCheckRefinementBlock()
 //! \brief Refinement criterion for disk pgen
 inline parthenon::AmrTag ProblemCheckRefinementBlock(MeshBlockData<Real> *mbd) {
-  PARTHENON_FAIL("Disk user-defined AMR criterion not yet implemented!");
+  PARTHENON_FAIL("Disk pgen-defined AMR criterion not yet implemented!");
   return AmrTag::same;
 }
 
--- a/src/artemis.hpp
+++ b/src/artemis.hpp
@@ -148,6 +148,8 @@ inline int ProblemDimension(parthenon::ParameterInput *pin) {
 
 namespace artemis {
 extern std::function<AmrTag(MeshBlockData<Real> *mbd)> ProblemCheckRefinementBlock;
+extern std::function<TaskStatus(MeshData<Real> *md, const Real time, const Real dt)>
+    ProblemGeneratorSourceTerm;
 } // namespace artemis
 
 #endif // ARTEMIS_ARTEMIS_HPP_
--- a/src/pgen/problem_modifier.hpp
+++ b/src/pgen/problem_modifier.hpp
@@ -28,9 +28,9 @@
 
 // User-defined refinement criterion callback
 namespace artemis {
-
 std::function<AmrTag(MeshBlockData<Real> *mbd)> ProblemCheckRefinementBlock = nullptr;
-
+std::function<TaskStatus(MeshData<Real> *md, const Real time, const Real dt)>
+    ProblemGeneratorSourceTerm = nullptr;
 } // namespace artemis
 
 // Problem modifiers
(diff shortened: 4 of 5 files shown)


src/artemis_driver.cpp
```diff
@@ -234,10 +234,15 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
                                 time, bdt);
       }
 
+      // Apply pgen-defined source term
+      // NOTE(@pdmullen): Intentional positioning after S(W) srcs and before S(U) srcs
```
brryan, line 238: What is `S(W)`?

brryan (commented):
Seems reasonable to me... is this ready to go now without any actual use in anticipation of (I assume) your radiating planet problem, or should we try to merge that in with this work as part of it?

src/artemis_driver.cpp
pdmullen: `S(W)`: source terms computed from (begin-stage) primitives:
`S(U)`: source terms computed from (partially-updated) conserved variables. 

for example in the 1st stage of Heun's method

U1* = U0  + dt divF[U0] + dt S[W]
U1 = U1* + dt S[U1*]

pdmullen:
> Seems reasonable to me... is this ready to go now without any actual use in anticipation of (I assume) your radiating planet problem, or should we try to merge that in with this work as part of it?

I think I'd prefer to not push the isolated planet setup.  Adam and I are discussing other pgen setups that better exploit some of `artemis`'s functionality.  If the functionality in this PR is too isolated and not more generally useful to the `artemis` community, I would be totally fine with killing this PR.

brryan:
> > Seems reasonable to me... is this ready to go now without any actual use in anticipation of (I assume) your radiating planet problem, or should we try to merge that in with this work as part of it?
> 
> I think I'd prefer to not push the isolated planet setup. Adam and I are discussing other pgen setups that better exploit some of `artemis`'s functionality. If the functionality in this PR too isolated and not more generally useful to the `artemis` community, I would be totally fine with killing this PR.

It seems to me that we want to provide a general problem-specific source function? It would just be nice to combine it with a use case so that it is, like, compiled during the CI to check that it works. Maybe we don't have an immediate idea of when such a source for a problem will be needed by us.

src/artemis_driver.cpp
adamdempsey90: We could clarify this and call it `ExplicitProblemSourceTerm`, but not a big deal.

adamdempsey90:
> > > Seems reasonable to me... is this ready to go now without any actual use in anticipation of (I assume) your radiating planet problem, or should we try to merge that in with this work as part of it?
> > 
> > 
> > I think I'd prefer to not push the isolated planet setup. Adam and I are discussing other pgen setups that better exploit some of `artemis`'s functionality. If the functionality in this PR too isolated and not more generally useful to the `artemis` community, I would be totally fine with killing this PR.
> 
> It seems to me that we want to provide a general problem-specific source function? It would just be nice to combine it with a use case so that it is, like, compiled during the CI to check that it works. Maybe we don't have an immediate idea of when such a source for a problem will be needed by us.

I agree it would be nice to have a working example to go with this change. I'm sure we can think of some minor source function for one of the existing test problems. Like maybe in the conduction problem, you add some heat source/sink. Or you implement some special damping function in the disk problem. Doesn't have to be something serious. But I don't feel strongly about needing something so long as you've tested separately that this works

pdmullen:
> > > Seems reasonable to me... is this ready to go now without any actual use in anticipation of (I assume) your radiating planet problem, or should we try to merge that in with this work as part of it?
> > 
> > 
> > I think I'd prefer to not push the isolated planet setup. Adam and I are discussing other pgen setups that better exploit some of `artemis`'s functionality. If the functionality in this PR too isolated and not more generally useful to the `artemis` community, I would be totally fine with killing this PR.
> 
> It seems to me that we want to provide a general problem-specific source function? It would just be nice to combine it with a use case so that it is, like, compiled during the CI to check that it works. Maybe we don't have an immediate idea of when such a source for a problem will be needed by us.

I think this feature might be a nicety for some folks and is a commonly provided utility in other astro-codes.  

Luckily the CI will compile the new task and its execution (but it will always trigger the `nullptr` pathway).  

I feel goofy adding in a no-op source term to a pgen for the sole purpose of checking that the other pathway triggers correctly.  Maybe we need a unique pgen dedicated to testing every possible custom enrollment, but it isn't actually a true pgen, moreso just a unit test --- but this is out-of-scope for this PR.

src/artemis_driver.cpp
pdmullen: The current positioning does permit you to make this the "first" S(U) source if you really like.

brryan: Chatted on the phone -- I was momentarily confused about meshdata registers vs prim/cons state vectors.

Also I think this can still be `ImplicitProblemSourceTerm` since we can't for example stop people from doing a backward Euler update inside this call, so I think keeping the name as-is is appropriate for now.

brryan (approved):
LGTM, good feature to have. Any objections to merge @adamdempsey90?

src/artemis_driver.cpp
adamdempsey90: Sure, if we were to split into an explicit and implicit version, it would just make it clear where they go in the driver and what they could do. I'm fine as is, of course

src/pgen/disk.hpp
```diff
@@ -814,11 +814,21 @@ void DiskBoundaryExtrap(std::shared_ptr<MeshBlockData<Real>> &mbd, bool coarse)
...
 
+//----------------------------------------------------------------------------------------
+//! \fn TaskStatus ProblemGeneratorSourceTerm()
+//! \brief Custom source term for disk pgen
+template <Coordinates GEOM>
+TaskStatus ProblemGeneratorSourceTerm(MeshData<Real> *md, const Real time,
+                                      const Real dt) {
+  PARTHENON_FAIL("Disk pgen-defined source term not yet implemented!");
+  return TaskStatus::complete;
+}
```
adamdempsey90, line 825: Is there a reason why you put a stub in this pgen and not all others? I'm not sure it's needed since it's clear from the `std::function` in `problem_modifier` what the signature is.

adamdempsey90:
> LGTM, good feature to have. Any objections to merge @adamdempsey90?

I would still like to see a modification of one of the existing tests with a super simple source term. @pdmullen is that out of the question right now?

src/pgen/disk.hpp
pdmullen: Just as a demonstration.  Notice that we also use the disk pgen to hold a stub for user-defined refinement criteria.

pdmullen:
> > LGTM, good feature to have. Any objections to merge @adamdempsey90?
> 
> I would still like to see a modification of one of the existing tests with a super simple source term. @pdmullen is that out of the question right now?

This is a totally reasonable request.  But I think the "right" solution here is introducing a `unit_pgen.cpp` that makes every possible custom enrollment that `artemis` supports (FYI user-defined refinement criteria are also not checked under CI currently).  I am happy to delay the merge of this MR until I implement something of the sort, if you'd like.

adamdempsey90:
> > > LGTM, good feature to have. Any objections to merge @adamdempsey90?
> > 
> > 
> > I would still like to see a modification of one of the existing tests with a super simple source term. @pdmullen is that out of the question right now?
> 
> This is a totally reasonable request. But I think the "right" solution here is introducing a `unit_pgen.cpp` that makes every possible custom enrollment that `artemis` supports (FYI user-defined refinement criteria are also not checked under CI currently). I am happy to delay the merge of this MR until I implement something of the sort, if you'd like.

I don't see the rush in merging this, so I'm fine with waiting until the CI covers this new stuff. You could add the user defined refinement criteria to your new test too ;)

PR #31, opened 2024-12-21
[Fix] Fixes bug in particle forces

adamdempsey90:
We were not properly saving the cumulative step force for stages that weren't the last stage. `rk1` and `vl2` integrators were fine, but `rk2` and `rk3` were wrong. 

This fixes that.

This plot shows the calculated torque with and without this fix. 
force_fix

## Checklist


brryan (approved):
Yes I agree with this change, nice catch.

PR #34, opened 2025-01-06
Updates for GM and nbody particles

adamdempsey90:
## Background

This does a few things:

- Does away with `mass_tot` in `<gravity>` and effectively makes the base `<gravity>` node optional. 

  `<gravity/point>` and `<gravity/binary>` have `mass` parameters that set their mass now.
  `<nbody>` now has an optional parameter that will rescale the entire system to the desired mass. 

  Because `<nbody>` can set `GM` now, the gravity package is initialized after nbody and takes the `Packages_t package` in its `Initialize` function so it can pull out nbody's `GM`. 

  I've also moved the printing of initial particle properties to the big artemis splash screen. Previously it was before this and looked weird. 

- Some fixes for how the shearing box interacts with "local" nbody systems (e.g., a single point mass at the center of the box). 

- Removes the atomic adds in nbody gravity. Those were left over from when the outer loop over particles was a `par_for`. 

- Fixes an issue with CGS units where you would have to specify the length, time, and mass factors explicitly. But I should be able to just say `physicsl_units = cgs` and that should mean real cgs and the conversion factors all default to 1. 

## Checklist

### DIFF
--- a/doc/src/physics.rst
+++ b/doc/src/physics.rst
@@ -673,10 +673,7 @@ Specifically, |code|, adds the body force,
 .. math::
    \rho \frac{D \mathbf{v}}{D t} = \rho \mathbf{g}
 
-To activate external gravity, there must be a ``<gravity>`` node and ``gravity = true`` under the ``<physics>`` node.
-The ``<gravity>`` requires setting the parameter ``gm``, typically this is set to ``gm = 1.0``.
-In addition to ``gm``, the ``tstart`` and ``tstop`` parameters control when gravity is active.
-
+To activate external gravity, there must be a subnode of ``<gravity>`` (e.g., ``<gravity/point>``) and ``gravity = true`` under the ``<physics>`` node.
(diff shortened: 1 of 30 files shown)


adamdempsey90:
TODO

- [x] Update input files for new mass parameters
- [x] Update docs

brryan:
>     * Fixes an issue with CGS units where you would have to specify the length, time, and mass factors explicitly. But I should be able to just say `physicsl_units = cgs` and that should mean real cgs and the conversion factors all default to 1.

I would call this a choice rather than an issue. "physical units" and "code units" are distinct ideas; if we want code units to always default to physical units (the change made here, where previously code units only defaulted to physical units if physical units were `scale_free`), I think that is fine.

adamdempsey90:
@pdmullen @brryan  this should be ready to go

pdmullen (approved):
All looks good to me!

PR #38, opened 2025-01-08
Testing issues

### LINKED ISSUE #35: Testing script and mpiexec
The testing scripts use `mpiexec` to launch artemis, but not all clusters have `mpiexec` available.  Venado uses `srun`, for example.

brryan:
## Background

A few small issues have cropped up in our testing. 

Fixes #35 

## Description of Changes

- [x] Always use `clang-format-12` on CI for formatting
- [x] Use `Agg` matplotlib backend if no display is set to avoid crashing when plotting test output
- [x] Allow for `srun` in test script if `mpiexec` is unavailable

## Checklist

### DIFF
--- a/tst/scripts/binary/binary.py
+++ b/tst/scripts/binary/binary.py
@@ -29,11 +29,11 @@ logger = logging.getLogger("artemis" + __name__[7:])  # set logger name
 logging.getLogger("h5py").setLevel(logging.WARNING)
 logging.getLogger("matplotlib").setLevel(logging.WARNING)
 import scripts.utils.analysis as analysis
-import matplotlib.colors as colors
-import matplotlib.pyplot as plt
 import matplotlib
 
 matplotlib.use("Agg")  # Use the Agg backend to avoid issues with DISPLAY not being set
+import matplotlib.colors as colors
+import matplotlib.pyplot as plt
 
 _nranks = 1
 _file_id = "binary"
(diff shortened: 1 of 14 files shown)


tst/scripts/diffusion/alpha_disk.py
```diff
@@ -27,6 +27,9 @@
 import scripts.utils.analysis as analysis
 import matplotlib.colors as colors
 import matplotlib.pyplot as plt
+import matplotlib
+if not os.getenv('DISPLAY'):
+    matplotlib.use('Agg')  # Use the Agg backend if no display is available
```
adamdempsey90: I honestly don't recall `run_test.py` ever popping up a plot when running it, so should we just always use Agg?

brryan: Yeah I don't think there's a significant reason to not always use `Agg` here. I'll make that change.

tst/scripts/binary/binary.py
```diff
@@ -31,6 +31,9 @@
 import scripts.utils.analysis as analysis
 import matplotlib.colors as colors
 import matplotlib.pyplot as plt
+import matplotlib
+
+matplotlib.use("Agg")  # Use the Agg backend to avoid issues with DISPLAY not being set
```
adamdempsey90, line 34: I was always under the impression that you had to put the matplotlib load and use statement before loading any submodule like pyplot? Maybe I'm wrong

brryan: Oops yeah you're right. I'm a bit surprised that passed the CI actually. Anyway, fixed.

adamdempsey90 (approved):

PR #39, opened 2025-01-09
Add some extra information about how artemis/parthenon are finishing up

adamdempsey90:
## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -119,7 +119,7 @@ int ArtemisFinalize(const DriverStatus status, const bool quick_exit,
     if (Globals::my_rank == 0) std::cout << "artemis driver timed out!" << std::endl;
     ret = 2;
   } else {
-    PARTHENON_WARN("artemis driver returned with an uknown code!");
+    PARTHENON_WARN("artemis driver returned with an unknown code!");
   }
   // Call MPI_Finalize and Kokkos::finalize if necessary
   // MPI and Kokkos can no longer be used
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -131,23 +131,25 @@ int main(int argc, char *argv[]) {
   // Run artemis
   auto status = LaunchWorkFlow(pman, pman.pinput.get());
 
+  int ret = 3;
+  if (status == DriverStatus::complete) {
+    std::cout << "artemis driver complete!" << std::endl;
+    ret = 0;
+  } else if (status == DriverStatus::failed) {
+    std::cout << "artemis driver failed!" << std::endl;
+    ret = 1;
+  } else if (status == DriverStatus::timeout) {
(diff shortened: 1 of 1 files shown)


brryan (approved):
This looks like a reasonable change to me although my suspicion is the issue we're seeing now causes the program to exit before `ParthenonFinalize()` returns so I suspect this won't help us dig into that issue.

adamdempsey90:
> This looks like a reasonable change to me although my suspicion is the issue we're seeing now causes the program to exit before `ParthenonFinalize()` returns so I suspect this won't help us dig into that issue.

Yes, all this adds is more information about where we are / what has finished at exit.

adamdempsey90:
> > This looks like a reasonable change to me although my suspicion is the issue we're seeing now causes the program to exit before `ParthenonFinalize()` returns so I suspect this won't help us dig into that issue.
> 
> Yes, all this adds is more information about where we are / what has finished at exit.

But, I've just added a new parameter, `artemis/quick_exit`, that will not call ParthenonFinalize at all. I assume this is dangerous, but do you think it's worth the risk to bypass a hang in there?

brryan:
> > > This looks like a reasonable change to me although my suspicion is the issue we're seeing now causes the program to exit before `ParthenonFinalize()` returns so I suspect this won't help us dig into that issue.
> > 
> > 
> > Yes, all this adds is more information about where we are / what has finished at exit.
> 
> But, I've just added a new parameter, `artemis/quick_exit`, that will not call ParthenonFinalize at all. I assume this is dangerous, but do you think it's worth the risk to bypass a hang in there?

Oh that's a good idea. I don't think it's that risky. Could solve this problem. 

Maybe the mature solution is to wrap `ParthenonFinalize` in some kind of exception handling? I'm not immediately sure if or how that would work with Kokkos and MPI operations.

src/main.cpp
```diff
@@ -105,6 +105,38 @@ parthenon::DriverStatus LaunchWorkFlow(parthenon::ParthenonManager &pman,
...
+    if (Globals::my_rank == 0) std::cout << "artemis driver complete!" << std::endl;
+    ret = 0;
+  } else if (status == DriverStatus::failed) {
+    if (Globals::my_rank == 0) std::cout << "artemis driver failed!" << std::endl;
+    ret = 1;
+  } else if (status == DriverStatus::timeout) {
+    if (Globals::my_rank == 0) std::cout << "artemis driver timed out!" << std::endl;
+    ret = 2;
+  } else {
+    PARTHENON_WARN("artemis driver returned with an uknown code!");
```
brryan:
```suggestion
    PARTHENON_WARN("artemis driver returned with an unknown code!");
```

adamdempsey90: that word is the bane of my existence

brryan: Mine is brocolli

PR #43, opened 2025-01-11
Convert dust params from cgs

### LINKED ISSUE #40: Input dust parameters in cgs and convert to internal units
This is now possible with the units machinery.

adamdempsey90:
## Background

This converts the input dust sizes and grain density from cgs to problem units. 

Closes #40 

## Checklist

### DIFF
--- a/src/dust/dust.cpp
+++ b/src/dust/dust.cpp
@@ -101,18 +101,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
 
   // Dust sizes
   const auto size_dist = pin->GetOrAddString("dust", "size_input", "direct");
-  const auto input_units = pin->GetOrAddString("dust", "input_units", "cgs");
-  Real length_conv = Null<Real>();
-  Real rho_conv = Null<Real>();
-  if (input_units == "cgs") {
-    length_conv = units.GetLengthPhysicalToCode();
-    rho_conv = units.GetMassDensityPhysicalToCode();
-  } else if (input_units == "problem") {
-    length_conv = 1.;
-    rho_conv = 1.;
-  } else {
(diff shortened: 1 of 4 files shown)


src/dust/dust.cpp
```diff
@@ -99,10 +101,23 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 
   // Dust sizes
   const auto size_dist = pin->GetOrAddString("dust", "size_input", "direct");
+  const auto input_units = pin->GetOrAddString("dust", "input_units", "cgs");
+  Real length_conv = Null<Real>();
+  Real rho_conv = Null<Real>();
+  if (input_units == "cgs") {
+    length_conv = units.GetLengthPhysicalToCode();
+    rho_conv = units.GetMassDensityPhysicalToCode();
+  } else if (input_units == "problem") {
```
brryan: Is this necessary? Can we just remove `input_units` and have PhysicalToCode convert them from cgs to physical units if our physical units are in CGS, and if the physical units are scalefree, length_conv etc. will just be `1` anyway?

adamdempsey90: test problems may not want these things input in cgs, so I provide that option.

brryan: I agree not all test problems will want cgs, but in that case won't those problems have `artemis/physical_units = scalefree` in which case `units.GetLengthPhysicalToCode() = units.GetMassDensityPhysicalToCode() = 1` anyway?

adamdempsey90: hmm, i suppose so. It feels weird to make assumptions like that, but I don't see a big problem with it

pdmullen (approved):
LGTM!

PR #44, opened 2025-01-12
Update singularity-eos

adamdempsey90:
## Background

So that we can use `MeanAtomicMass()` to pull out `mu` without having to pass it and `amu` to functions we are already passing the eos to. 

I also want to see if `jaybenne` will build properly...because I'm having problems on another branch when updating singularity.

## Checklist

- [ ] New features are documented
- [ ] Tests added for bug fixes and new features
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #45, opened 2025-01-13
Draft: Refactor stopping times

### LINKED ISSUE #41: Expand stopping time
Following https://iopscience.iop.org/article/10.1088/0004-637X/806/2/203 

At the same time, we should also follow the Riemann solvers / diffusion coefficients and make the stopping times template specialized classes

adamdempsey90:
## Background

This does two things.

1. Refactors our stopping times to be specialized classes following the diffusion coefficients.
2. Adds a more sophisticated model from https://iopscience.iop.org/article/10.1088/0004-637X/806/2/203 

This relies on #44 . 

Closes #41 .

## Checklist


adamdempsey90:
TODO:

- [ ] Add new test for new model

src/drag/drag.hpp
```diff
@@ -137,13 +139,124 @@ struct StoppingTimeParams {
         h_tau(n) = scale;
       }
       tau.DeepCopy(h_tau);
+    } else if (choice == "dp15") {
+      model = DragModel::dp15;
+      scale = pin->GetOrAddReal(block_name, "scale", 1.0);
```
pdmullen, line 144: does this need a time unit conversion or something?

src/drag/drag.hpp
```diff
@@ -137,13 +139,124 @@ struct StoppingTimeParams {
...
+      auto h_tau = tau.GetHostMirror();
+      for (int n = 0; n < nd; n++) {
+        h_tau(n) = scale;
+      }
+      tau.DeepCopy(h_tau);
+      dh = units.GetLengthPhysicalToCode() *
+           pin->GetOrAddReal(block_name, "gas_diameter", 2.71e-8 /* cm */);
+      p1 = pin->GetOrAddReal(block_name, "p1", 3.07);
```
pdmullen, line 152: dimensionless?

src/drag/drag.hpp
```diff
@@ -137,13 +139,124 @@ struct StoppingTimeParams {
...
+template <>
+class DragCoeff<DragModel::dp15> {
+ public:
+  KOKKOS_INLINE_FUNCTION Real Get(const StoppingTimeParams &dp, const int id,
+                                  const Real dg, const Real Tg, const Real u,
+                                  const Real grain_density, const Real size,
+                                  const EOS &eos) const {
+    const Real cv = eos.SpecificHeatFromDensityTemperature(dg, Tg);
+    const Real gm1 = eos.GruneisenParamFromDensityTemperature(dg, Tg);
+    const Real mu = dp.mass_scale * eos.MeanAtomicMass();
```
pdmullen, line 245: A few thoughts here... 

I think what we really want here is the mean molecular weight `mu` which is not the same as the mean atomic mass.  For mean molecular weight, I don't think you can really gather that from what singularity provides...

Also I have a question about units.  What is singularity returning here for `MeanAtomicMass`? As written, it appears it is returning something in CGS (as opposed to, e.g., amu).  I think that our singularity invocations in the future may already have knowledge of the unit system `artemis` is employing (didn't we have a discussion about this during one of our group meetings @brryan?).

src/drag/drag.hpp
adamdempsey90: yes

src/drag/drag.hpp
adamdempsey90: scale is just a scale on the overall stopping time, so no unit conversion needed

src/drag/drag.hpp
adamdempsey90: I had the same question about units w.r.t singularity-eos. I don't think we do the same "NonCGS" units that we do for opac. 

And yeah, I'm missing the unit conversion here. 

I looked more carefully at `IdealGas` and I believe we can do what I want here -- we just have to pass `MeanAtomicProperties(mu, zbar)` to the eos along with a separately calculated cv when we initialize it. Then this function call here will return that `mu` we initially passed. So I think it can be made to work.

Really, though, `IdealGas` could be setting `cv` internally from the given `abar`, or setting `abar` from the given `cv` since `cv = kb/(abar * amu * (g-1))` (or `mh` instead of `amu`).  That seems like something that should be fixed in singularity at some point.

adamdempsey90: And just to be clear, at the level of physics that we currently handle there is no functional difference between mean atomic mass and mean molecular weight. For example, if we are just hydrogen, then `abar = 1.008 amu`, if we are H2 then `abar = 2*1.008 amu`. 

Maybe this is an abuse of the system, but I think it helps reduce the number of things we're passing around (by storing `mu` in the eos).

src/drag/drag.hpp
pdmullen: Thanks!

src/drag/drag.hpp
pdmullen: thanks!

src/drag/drag.hpp
adamdempsey90: As discussed offline, we'll overload the term abar to mean both mean atomic mass and mean molecular weight for now until singularity provides mmw and/or artemis requires the distinction. 

The latest commits fix the initialization of abar/zbar in the eos and adds amu to the `mass_scale` variable, so the units are correct now.

pdmullen: Understood to be a bit of a hack... okay with it for now.

pdmullen (approved):

src/drag/drag.hpp
```diff
@@ -137,13 +139,125 @@ struct StoppingTimeParams {
...
+class DragCoeff<DragModel::stokes> {
+ public:
+  KOKKOS_INLINE_FUNCTION Real Get(const StoppingTimeParams &dp, const int id,
+                                  const Real dg, const Real Tg, const Real u,
+                                  const Real grain_density, const Real size,
+                                  const EOS &eos) const {
+    const Real cv = eos.SpecificHeatFromDensityTemperature(dg, Tg);
+    const Real gm1 = eos.GruneisenParamFromDensityTemperature(dg, Tg);
+    //  kb T/ mu = cv * gm1 * T
+    const Real vth = std::sqrt(8.0 / M_PI * gm1 * cv * Tg);
```
adamdempsey90, line 231: Now that we have constants, we can just directly use kb/mu*T

pdmullen:
Why did this never get merged?

adamdempsey90:
> Why did this never get merged?

The plan was to do some testing

111-1000:
In src/drag/drag.hpp, in 

const Real CEu = 1. / (sgam * Mu) * (4.6 / (1. + M) + 1.7 /* srt(Td/Tg) */);

we may want to apply the following correction:

4.6 -> (8/3)*sqrt(8/pi)~~4.25538612146777
1.7 -> (pi/3)*sqrt(8/pi)~~1.67108481066774

which would provide the original Epstein drag limit. I was aware of the 4.6-4.26 discrepancy in Melosh & Goldin paper, but at the time, I decided to leave it and defer for later (I was working on planetesimals, so Epstein regime was irrelevant). I still don't know if it was a typo or something else. But I think we are OK reverting to the original Epstein formulation.

pdmullen:
I reviewed this one before, but it looks like it never made it into`develop` because we were planning additional testing.  We are quite behind `develop` now though.  So I am changing to Draft.

PR #46, opened 2025-01-18
Fix rebound on mac

adamdempsey90:
## Background

This should fix the linking issues on mac. Essentially we don't try in cmake, and instead just run  `install_name_tool` after the build completes to set the `@rpath` in the binary directly.  Because of this change, we also no longer need to move `librebound.so` after building it, so I got rid of that step too.

With this, I can now successfully execute artemis from anywhere and move the binary anywhere and execute it. 

I also added the `-Wno-deprecated-declarations` to suppress the warnings from singularity. 

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,11 @@ if(ARTEMIS_ENABLE_ASAN)
   add_link_options(-fsanitize=address -fsanitize=undefined)
 endif()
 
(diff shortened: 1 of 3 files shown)


pdmullen (approved):
Just tested this locally on my mac (via `disk_collisions.in`) and it worked! Nice find (albeit I have no idea what this new `cmake` nonsense is actually doing...)

PR #47, opened 2025-01-18
Add temperature to units and have units aware EOS

adamdempsey90:
## Background

This does a few things:
-  adds the missing temperature specifier for our units. That shouldn't change anything. 
-  Makes singularity-eos aware of our unit system so that it's on par with singularity-opac.

## Checklist

### DIFF
--- a/src/utils/units.cpp
+++ b/src/utils/units.cpp
@@ -32,25 +32,29 @@ Units::Units(ParameterInput *pin, std::shared_ptr<StateDescriptor> pkg) {
   } else {
     PARTHENON_FAIL("Physical unit system not recognized! Choices are [scalefree, cgs]");
   }
-
-  if (physical_units_ == PhysicalUnits::scalefree) {
-    length_ = 1.;
-    time_ = 1.;
-    mass_ = 1.;
-  } else {
+  length_ = 1.;
+  time_ = 1.;
+  mass_ = 1.;
+  temp_ = 1.;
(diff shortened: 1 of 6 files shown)


src/utils/units.cpp
```diff
@@ -32,25 +32,29 @@ Units::Units(ParameterInput *pin, std::shared_ptr<StateDescriptor> pkg) {
...
+    if (unit_conversion == "ppd") {
       length_ = AU;
       mass_ = Msolar;
       time_ = Year / (2. * M_PI);
+    } else if (unit_conversion == "base") {
+      // do nothing
     } else {
       PARTHENON_FAIL("Unit conversion not recognized! Choices are [base, ppd]");
     }
+    // not that these multiplied by whatever values were previously set.
```
pdmullen: I had engrained in my head that `length`, `time`, `mass` should only ever be used with `unit_conversion = base` and should always be passed in `cgs` units.  

I understand that this change does address @shengtai's concern from last meeting, but it does feel a bit mistake-prone without clear documentation.  In that vein, if you really like this model (rather than just demanding users to pass e.g., `mass = 1.989e34` for a more "custom" unit system), can you please document this? Also, if you're okay with the style choice: I'd prefer to shuffle things around in this `if`/`else` construct so that there isn't a bare `else if`.

adamdempsey90: As discussed offline, I've reverted this part to how it was before.

pdmullen (approved):

PR #55, opened 2025-02-06
BugFix Temperature units in disk

### LINKED ISSUE #53: TempProfile in disk pgen needs units

### LINKED ISSUE #54: -O3 in releases
Our "release" build is actually `RelWithDebInfo` which only has `-O2`. We should change that to be `-O3`.

adamdempsey90:
## Background

- Add the correct `kb/mu` factors to the disk temperature profile.
- Allow constant viscosity
- Force `-O3` in `RelWithDebInfo` builds

Closes #53 
Closes #54 

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,10 @@ option(ARTEMIS_ENABLE_OPENMP "Enable OpenMP for artemis and parthenon" OFF)
 option(ARTEMIS_ENABLE_COMPILE_TIMING "Enable timing of compilation of artemis" ON)
 option(ARTEMIS_ENABLE_ASAN "Enable AddressSanitizer to detect memory errors" OFF)
 
+# Force -03 optimization for RelWithDebInfo
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG" CACHE STRING "Force -O3 in RelWithDebInfo mode" FORCE)
+set(CMAKE_C_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG" CACHE STRING "Force -O3 in RelWithDebInfo mode" FORCE)
+
 # Timing of compilation
 if (ARTEMIS_ENABLE_COMPILE_TIMING)
   set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE
--- a/src/pgen/disk.hpp
+++ b/src/pgen/disk.hpp
@@ -39,6 +39,7 @@
 #include "nbody/nbody.hpp"
 #include "utils/artemis_utils.hpp"
 #include "utils/eos/eos.hpp"
+#include "utils/units.hpp"
 
 using ArtemisUtils::EOS;
 using ArtemisUtils::VI;
@@ -60,6 +61,7 @@ struct DiskParams {
   Real alpha, nu0, nu_indx;
   Real mdot;
   Real temp_soft2;
+  Real kbmu;
   bool do_dust;
   bool nbody_temp;
   bool quiet_start;
@@ -101,7 +103,8 @@ Real TempProfile(struct DiskParams pgen, const Real R, const Real z) {
   const Real H = R * pgen.h0 * std::pow(R / pgen.r0, pgen.flare);
   const Real ir1 = 1.0 / std::sqrt(R * R + pgen.temp_soft2);
   const Real omk2 = SQR(pgen.Omega0) * ir1 * ir1 * ir1;
-  const Real T0 = omk2 * H * H / pgen.Gamma;
+  // c_iso^2 = P/rho = kb/mu T = Omk^2 H^2
+  const Real T0 = omk2 * H * H / (pgen.kbmu * pgen.Gamma);
   return T0 * std::pow(rho / rho0, pgen.Gamma - 1.0);
 }
 
@@ -275,6 +278,9 @@ inline void InitDiskParams(MeshBlock *pmb, ParameterInput *pin) {
     disk_params.l0 = pin->GetOrAddReal("problem", "l0", 0.0);
     disk_params.dust_to_gas = pin->GetOrAddReal("problem", "dust_to_gas", 0.01);
     disk_params.temp_soft2 = pin->GetOrAddReal("problem", "temp_soft", 0.0);
+    const auto mu = gas_pkg->Param<Real>("mu");
+    auto &constants = artemis_pkg->Param<ArtemisUtils::Constants>("constants");
+    disk_params.kbmu = constants.GetKBCode() / (mu * constants.GetAMUCode());
 
     disk_params.do_dust = params.Get<bool>("do_dust");
 
@@ -312,7 +318,7 @@ inline void InitDiskParams(MeshBlock *pmb, ParameterInput *pin) {
         disk_params.nu0 = disk_params.alpha * disk_params.gamma_gas *
                           SQR(disk_params.h0 * disk_params.r0 * disk_params.Omega0);
         disk_params.nu_indx = 1.5 + disk_params.q;
-      } else if (vtype == "powerlaw") {
+      } else if ((vtype == "powerlaw") || (vtype == "constant")) {
         disk_params.nu0 = pin->GetReal("gas/viscosity", "nu");
         disk_params.nu_indx = pin->GetOrAddReal("gas/viscosity", "r_exp", 0.0);
       } else {


adamdempsey90:
Note also that for the cooling package, you have to include the correct `kb/mu` in you input file if you previously defined it as `P/rho`

src/pgen/disk.hpp
```diff
@@ -275,6 +278,9 @@ inline void InitDiskParams(MeshBlock *pmb, ParameterInput *pin) {
     disk_params.l0 = pin->GetOrAddReal("problem", "l0", 0.0);
     disk_params.dust_to_gas = pin->GetOrAddReal("problem", "dust_to_gas", 0.01);
     disk_params.temp_soft2 = pin->GetOrAddReal("problem", "temp_soft", 0.0);
+    const auto mu = gas_pkg->Param<Real>("mu");
+    auto &constants = artemis_pkg->Param<ArtemisUtils::Constants>("constants");
+    disk_params.kbmu = constants.GetKBCode() / (mu * constants.GetAMUCode());
```
pdmullen, line 283: Can we rename this to indicate that it also includes a factor of amu? I usually think of mu itself as being dimensionless.

src/pgen/disk.hpp
```diff
@@ -101,7 +103,8 @@ Real TempProfile(struct DiskParams pgen, const Real R, const Real z) {
   const Real H = R * pgen.h0 * std::pow(R / pgen.r0, pgen.flare);
   const Real ir1 = 1.0 / std::sqrt(R * R + pgen.temp_soft2);
   const Real omk2 = SQR(pgen.Omega0) * ir1 * ir1 * ir1;
-  const Real T0 = omk2 * H * H / pgen.Gamma;
+  // c_iso^2 = P/rho = kb/mu T = Omk^2 H^2
+  const Real T0 = omk2 * H * H / (pgen.kbmu * pgen.Gamma);
```
pdmullen, line 107: I trust that you implemented this correctly... but just for my own understanding:

Could we have equivalently kept everything (mostly) as was, working in this pgen in a "scale free capacity"... then only at the end introduce a conversion to cgs, then to artemis code units, _then_ doing 

```
v(b, gas::prim::density(), k, j, i) = gdens * unit_conv;
v(b, gas::prim::sie(), k, j, i) = gsie * unit_conv;
```

This is just me trying to forward think about how people will interface with our units infrastructure and some potential pitfalls.

CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ option(ARTEMIS_ENABLE_OPENMP "Enable OpenMP for artemis and parthenon" OFF)
 option(ARTEMIS_ENABLE_COMPILE_TIMING "Enable timing of compilation of artemis" ON)
 option(ARTEMIS_ENABLE_ASAN "Enable AddressSanitizer to detect memory errors" OFF)
 
+# Force -03 optimization for RelWithDebInfo
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG" CACHE STRING "Force -O3 in RelWithDebInfo mode" FORCE)
```
pdmullen, line 29: Is `-O3` always safe? I wonder if there is a good reason why parth doesn't use it by default for release.  Maybe it is they avoid it due to extended compile times and/or memory useage?

pdmullen (approved):

shengtai:
I agree that “disk.hpp” pgen should be scale-free, even with user-input cgs-unit conversion.

From: Patrick Mullen ***@***.***>
Reply-To: lanl/artemis ***@***.***>
Date: Thursday, February 6, 2025 at 3:17 PM
To: lanl/artemis ***@***.***>
Cc: "Li, Shengtai" ***@***.***>, Review requested ***@***.***>
Subject: [EXTERNAL] Re: [lanl/artemis] BugFix Temperature units in disk (PR #55)


@pdmullen approved this pull request.

________________________________

In src/pgen/disk.hpp<https://urldefense.com/v3/__https:/github.com/lanl/artemis/pull/55*discussion_r1945546781__;Iw!!Bt8fGhp8LhKGRg!GB2gXrW2KFBi-83WTXkJsVAe4N1Fvr47KOL4JcDvzqMLXoCX2AqBYVwj7QoiaekDLqSSpoU-VWdQvYRWX7LWkA$>:

> @@ -275,6 +278,9 @@ inline void InitDiskParams(MeshBlock *pmb, ParameterInput *pin) {

     disk_params.l0 = pin->GetOrAddReal("problem", "l0", 0.0);

     disk_params.dust_to_gas = pin->GetOrAddReal("problem", "dust_to_gas", 0.01);

     disk_params.temp_soft2 = pin->GetOrAddReal("problem", "temp_soft", 0.0);

+    const auto mu = gas_pkg->Param<Real>("mu");

+    auto &constants = artemis_pkg->Param<ArtemisUtils::Constants>("constants");

+    disk_params.kbmu = constants.GetKBCode() / (mu * constants.GetAMUCode());

Can we rename this to indicate that it also includes a factor of amu? I usually think of mu itself as being dimensionless.

________________________________

In src/pgen/disk.hpp<https://urldefense.com/v3/__https:/github.com/lanl/artemis/pull/55*discussion_r1945549632__;Iw!!Bt8fGhp8LhKGRg!GB2gXrW2KFBi-83WTXkJsVAe4N1Fvr47KOL4JcDvzqMLXoCX2AqBYVwj7QoiaekDLqSSpoU-VWdQvYSZVBLhDA$>:

> @@ -101,7 +103,8 @@ Real TempProfile(struct DiskParams pgen, const Real R, const Real z) {

   const Real H = R * pgen.h0 * std::pow(R / pgen.r0, pgen.flare);

   const Real ir1 = 1.0 / std::sqrt(R * R + pgen.temp_soft2);

   const Real omk2 = SQR(pgen.Omega0) * ir1 * ir1 * ir1;

-  const Real T0 = omk2 * H * H / pgen.Gamma;

+  // c_iso^2 = P/rho = kb/mu T = Omk^2 H^2

+  const Real T0 = omk2 * H * H / (pgen.kbmu * pgen.Gamma);

I trust that you implemented this correctly... but just for my own understanding:

Could we have equivalently kept everything (mostly) as was, working in this pgen in a "scale free capacity"... then only at the end introduce a conversion to cgs, then to artemis code units, then doing

v(b, gas::prim::density(), k, j, i) = gdens * unit_conv;

v(b, gas::prim::sie(), k, j, i) = gsie * unit_conv;

This is just me trying to forward think about how people will interface with our units infrastructure and some potential pitfalls.

________________________________

In CMakeLists.txt<https://urldefense.com/v3/__https:/github.com/lanl/artemis/pull/55*discussion_r1945552088__;Iw!!Bt8fGhp8LhKGRg!GB2gXrW2KFBi-83WTXkJsVAe4N1Fvr47KOL4JcDvzqMLXoCX2AqBYVwj7QoiaekDLqSSpoU-VWdQvYTPg2EuRQ$>:

> @@ -25,6 +25,10 @@ option(ARTEMIS_ENABLE_OPENMP "Enable OpenMP for artemis and parthenon" OFF)

 option(ARTEMIS_ENABLE_COMPILE_TIMING "Enable timing of compilation of artemis" ON)

 option(ARTEMIS_ENABLE_ASAN "Enable AddressSanitizer to detect memory errors" OFF)



+# Force -03 optimization for RelWithDebInfo

+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG" CACHE STRING "Force -O3 in RelWithDebInfo mode" FORCE)

Is -O3 always safe? I wonder if there is a good reason why parth doesn't use it by default for release. Maybe it is they avoid it due to extended compile times and/or memory useage?

src/pgen/disk.hpp
adamdempsey90: sounds annoying to read. I actually consider `mu` to be the full thing btw.  How about I just change this to `pgen.temp_to_vel2`?

src/pgen/disk.hpp
adamdempsey90: I would prefer we have things called "temp" with units of temperature.

src/pgen/disk.hpp
pdmullen: We can keep it as is...

src/pgen/disk.hpp
pdmullen: Not arguing against your implementation. Just checking my knowledge. I remember in a similar circumstance in AthenaK, I renamed the variable something like pgas_over_rho

adamdempsey90: There isn't really any unit converting that needs to go on anywhere in this file except when dealing with temperature (where we were treating it as kb*T). We only really use that temperature to compute the pressure from a `P(rho,T)` eos call or an `E(rho,t)` call to set the final sie. We do that in several places, so I think putting taking out the kb/mu factor to recover true temperature at all those calling sites is not great. Instead it seems better to me to remove that kb/mu factor right away and have TempProfile return a temperature.

PR #57, opened 2025-02-24
Grey moment based radiation

adamdempsey90:
## Background

Adds a subcycled moment solver for grey radiation. 

Initial testing of a beam, thermalization, and the radiation shock looks good.

There are likely still some things to clean up (updating singularity for example). But I think this is ready for review. 

## Checklist


src/utils/fluxes/riemann/hlle.hpp
```diff
@@ -164,6 +174,32 @@ class RiemannSolver<RSolver::hlle, FLUID_TYPE> {
             } else if constexpr (FLUID_TYPE == Fluid::dust) {
               sl = std::min(wroe_ivx, wl_ivx);
               sr = std::max(wroe_ivx, wr_ivx);
+            } else if constexpr (is_grey<FLUID_TYPE>()) {
+              wl_ivx = (std::abs(wl_ivx) <= 1e-20) ? 0.0 : wl_ivx;
+              wl_ivy = (std::abs(wl_ivy) <= 1e-20) ? 0.0 : wl_ivy;
+              wl_ivz = (std::abs(wl_ivz) <= 1e-20) ? 0.0 : wl_ivz;
+              wr_ivx = (std::abs(wr_ivx) <= 1e-20) ? 0.0 : wr_ivx;
+              wr_ivy = (std::abs(wr_ivy) <= 1e-20) ? 0.0 : wr_ivy;
+              wr_ivz = (std::abs(wr_ivz) <= 1e-20) ? 0.0 : wr_ivz;
```
adamdempsey90: This feels like a hack

src/artemis.cpp
```diff
@@ -70,6 +72,13 @@ Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
   const bool do_viscosity = pin->GetOrAddBoolean("physics", "viscosity", false);
   const bool do_conduction = pin->GetOrAddBoolean("physics", "conduction", false);
   const bool do_radiation = pin->GetOrAddBoolean("physics", "radiation", false);
+  // Check if we are using IMC or moment based radiation
+  // This maybe is not great if user doesn't comment out unused node...
+  const bool do_imc = do_radiation && pin->DoesBlockExist("jaybenne");
+  const bool do_moment = do_radiation && pin->DoesBlockExist("radiation");
+  PARTHENON_REQUIRE(!(do_imc && do_moment),
+                    "You cannot have both a <jaybenne> block and a <radiation> block");
```
adamdempsey90: I think it would be better to have the input file as:
```
<radiation>
# common radiation params
<radiation/imc>
# jaybenne params
<radiation/moment>
# moment params
```

But that requires https://github.com/lanl/jaybenne/pull/1

src/artemis_driver.cpp
```diff
@@ -272,6 +279,123 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+    auto &tl = init_region[i];
+    auto &u0 = pmesh->mesh_data.GetOrAdd("u0", i);
+    auto &u1 = pmesh->mesh_data.GetOrAdd("u1", i);
+    tl.AddTask(none, ArtemisUtils::DeepCopyConservedData, u1.get(), u0.get());
+  }
+
+  const int nstages = rad_stages;
+  const std::array<Real, 2> beta{1.0, 0.5};
+  const std::array<Real, 2> gam0{0.0, 0.5};
+  const std::array<Real, 2> gam1{1.0, 0.5};
```
adamdempsey90: These end up being the same weights as `rk2`, so maybe I should just make a new rk2 integrator object

adamdempsey90: Fixed this so that we make a real integrator that corresponds to the rk1,rk2, and rk3 integrators. These are also the IMEX PD-ARS1/2/3 weights.

src/utils/fluxes/fluid_fluxes.hpp
```diff
@@ -198,8 +207,8 @@ TaskStatus CalculateFluxesImpl(MeshData<Real> *md, PKG &pkg, PackPrim vprim,
             // compute fluxes over [ks,ke+1]
             if (k > kl) {
               RiemannSolver<RIEMANN, FLUID_TYPE> riemann;
-              riemann.solve(eos, mbr, b, k, j, il, iu, X3DIR, wl, wr, vprim, vflux,
-                            vface);
+              riemann.solve(eos, c, chat, mbr, b, k, j, il, iu, X3DIR, wl, wr, vprim,
```
adamdempsey90: Maybe these extra optional arguments `(eos, c, chat)` can be packed up into a struct.

src/utils/integrators/artemis_integrator.hpp
```diff
@@ -50,24 +50,66 @@ inline TaskStatus DeepCopyConservedData(MeshData<Real> *to, MeshData<Real> *from
...
-
   // Extract integrator weights
   const Real gam0 = integrator->gam0[stage - 1];
   const Real gam1 = integrator->gam1[stage - 1];
   const Real beta_dt = integrator->beta[stage - 1] * integrator->dt;
 
-  // Packing and indexing
-  std::vector<MetadataFlag> flags({Metadata::Conserved, Metadata::WithFluxes});
-  static auto desc = MakePackDescriptor<any>(u0, flags, {parthenon::PDOpt::WithFluxes});
+  static auto desc =
```
pdmullen, line 97: This moves away from the abstraction that we originally enjoyed with `artemis`.  I'd propose we either (a) work with `MeshData` registers that do not carry the radiation variables when they are not needed or (b) specifically exclude fields with e.g., `MetadataRadiation`.  Haven't looked closely enough above to see if (a) is a viable strategy or not.

src/utils/integrators/artemis_integrator.hpp
```diff
@@ -50,24 +50,66 @@ inline TaskStatus DeepCopyConservedData(MeshData<Real> *to, MeshData<Real> *from
...
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, PARTHENON_AUTO_LABEL, parthenon::DevExecSpace(), 0,
+      md->NumBlocks() - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int &b, const int &k, const int &j, const int &i) {
+        for (int n = vf.GetLowerBound(b); n <= vf.GetUpperBound(b); ++n) {
+          vf.flux(b, X1DIR, n, k, j, i) = 0.0;
+          vf.flux(b, X1DIR, n, k, j, i) = 0.0;
+          vf.flux(b, X3DIR, n, k, j, i) = 0.0;
+        }
+        if (i == ib.e) {
```
pdmullen: this contains many loops over `n` and a lot of branching.  Would something like this

```c++
          vf.flux(b, X1DIR, n, k, j, i) = 0.0;
          vf.flux(b, X1DIR, n, k, j, i) = 0.0;
          vf.flux(b, X3DIR, n, k, j, i) = 0.0;
          vf.flux(b, X1DIR, n, k, j, i + (i == ib.e)) = 0.0;
          vf.flux(b, X2DIR, n, k, j + ((j == jb.e) && (multi_d)), i) = 0.0;
          vf.flux(b, X3DIR, n, k + ((k == kb.e) && (three_d)), j, i) = 0.0;
```

work?

src/utils/fluxes/riemann/llf.hpp
```diff
@@ -107,13 +111,40 @@ class RiemannSolver<RSolver::llf, FLUID_TYPE> {
               wr_ise = wr(ISE, i);
             }
 
+            Real scalel = 1.0;
```
pdmullen: Part of me feels like nestling all of this inside `llf.hpp` is a bit forced and maybe a bit confusing...  (e.g., the use of `wl_idn`)... Should we have a unique file/implementation of `llf` for radiation? In that scenario, the variable names could be more descriptive.

src/utils/fluxes/riemann/hlle.hpp
```diff
@@ -201,6 +237,11 @@ class RiemannSolver<RSolver::hlle, FLUID_TYPE> {
             if constexpr (FLUID_TYPE == Fluid::gas) {
               p.flux(b, dir, IPR, k, j, i) =
                   0.5 * (wl_ipr + wr_ipr) + qa * (wl_ipr - wr_ipr);
+            } else if constexpr (is_grey<FLUID_TYPE>()) {
+              wl_ipr = pscalel * wl_idn; // - sign?
```
pdmullen: the comment here scares me?

src/utils/fluxes/riemann/hlle.hpp
pdmullen: I agree.  Can you think of any other approaches that work in practice? If we adopt this current choice, then the units should probably make an appearance.

src/utils/fluxes/riemann/hlle.hpp
```diff
@@ -164,6 +174,32 @@ class RiemannSolver<RSolver::hlle, FLUID_TYPE> {
             } else if constexpr (FLUID_TYPE == Fluid::dust) {
               sl = std::min(wroe_ivx, wl_ivx);
               sr = std::max(wroe_ivx, wr_ivx);
+            } else if constexpr (is_grey<FLUID_TYPE>()) {
```
pdmullen: Same comments as in `llf.hpp` (below) regarding how this feels a bit forced... I might prefer a different set of Riemann solver for radiation.

src/utils/fluxes/fluid_fluxes.hpp
```diff
@@ -390,11 +406,33 @@ TaskStatus FluxSourceImpl(MeshData<Real> *md, PKG &pkg, PackPrim vprim, PackCons
...
           }
 
           // Add coordinate source term
-          const Real &dens = vprim(b, n, k, j, i);
-          const Real rdt = dens * dt;
+          Real rdt = vprim(b, n, k, j, i) * dt;
+          const Real cfac_ = c * chat;
+          if constexpr (is_grey<FLUID_TYPE>()) {
+            const Real f2 =
+                std::sqrt(SQR(vprim(b, IVX, k, j, i)) + SQR(vprim(b, IVY, k, j, i)) +
```
pdmullen, line 446: I'd prefer to introduce new `IER`, `IFX`, `IFY`, and `IFZ` (or similar) for radiation accesses.

src/radiation/moment/radiation.hpp
```diff
@@ -0,0 +1,180 @@
...
+    for (auto const &pmb : pmesh->block_list) {
+      const auto &reg = pmb->block_size;
+      for (int d = 0; d < pmesh->ndim; d++) {
+        const Real dx = (reg.xmax_[d] - reg.xmin_[d]) / reg.nx_[d];
+        dxmin = std::min(dxmin, dx);
+      }
+    }
+  }
+
+#ifdef MPI_PARALLEL
```
pdmullen, line 134: What's going on here? I doubt it will work with partitioning (albeit there are known other places in `artemis` that are not compatible with partitioning).

src/radiation/moment/radiation.hpp
```diff
@@ -0,0 +1,180 @@
...
+  const Real nx2 = fx2 / (f + Fuzz<Real>());
+  const Real nx3 = fx3 / (f + Fuzz<Real>());
+  f = std::min(1.0, f);
+
+  return {nx1 * f, nx2 * f, nx3 * f};
+}
+
+template <Coordinates GEOM>
+Real EstimateTimeStep(parthenon::Mesh *pmesh) {
+  // TODO:
```
pdmullen: Spooky comment.  Still relevant?

src/radiation/moment/radiation.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    // Check the type of refinement (e.g., gradient vs magnitude)
+    const std::string refine_type = pin->GetString("radiation", "refine_type");
+    const bool ref_grad = (refine_type == "gradient");
+    const bool ref_mag = (refine_type == "magnitude");
+    PARTHENON_REQUIRE((ref_grad || ref_mag) && !(ref_grad && ref_mag),
+                      "Only gradient or magnitude based criterion currently supported!");
+
+    // Specify appropriate AMR criterion callback
+    if (ref_grad) {
```
pdmullen, line 208: I don't think we need to support these for now.

src/radiation/moment/radiation.cpp
```diff
@@ -0,0 +1,494 @@
...
+        const auto &dx = coords.GetCellWidths();
+        Real denom = 0.0;
+        for (int d = 0; d < ndim; d++) {
+          denom += chat / dx[d];
+        }
+        ldt = std::min(ldt, 1.0 / denom);
+      },
+      Kokkos::Min<Real>(min_dt));
+
+  const auto cfl_number = params.template Get<Real>("cfl");
```
pdmullen: then what was the other `EstimateTimestep` I saw below?

src/radiation/moment/radiation.cpp
```diff
@@ -0,0 +1,494 @@
...
+          Real divf = (ax1[0] * v0.flux(b, X1DIR, n, k, j, i) -
+                       ax1[1] * v0.flux(b, X1DIR, n, k, j, i + 1));
+          if (multi_d)
+            divf += (ax2[0] * v0.flux(b, X2DIR, n, k, j, i) -
+                     ax2[1] * v0.flux(b, X2DIR, n, k, j + 1, i));
+          if (three_d)
+            divf += (ax3[0] * v0.flux(b, X3DIR, n, k, j, i) -
+                     ax3[1] * v0.flux(b, X3DIR, n, k + 1, j, i));
+          // Apply update
+          v0(b, n, k, j, i) =
```
pdmullen, line 384: I am missing why this structure doesn't nestle in with our `ApplyUpdate` in `utils`.

src/pgen/beam.hpp
```diff
@@ -0,0 +1,228 @@
...
+  auto &pars = artemis_pkg->Param<BeamParams>("beam_params");
+  pmb->par_for_bndry(
+      "BeamInnerX2", nb, IndexDomain::inner_x2, parthenon::TopologicalElement::CC, coarse,
+      fine, KOKKOS_LAMBDA(const int &l, const int &k, const int &j, const int &i) {
+        // Extract coordinates
+        geometry::Coords<GEOM> coords(pco, k, j, i);
+        const Real xf = coords.bnds.x1[0];
+
+        if (do_gas) {
+          for (int d = 0; d < 3; d++)
```
pdmullen: Can you wrap for loop in braces or just explicity set each three components?

src/derived/fill_derived.cpp
```diff
@@ -273,6 +310,27 @@ void PrimToCons(T *md) {
             mom3 = w_d * vel3 * hx[2];
           }
         }
+        if (do_rad) {
+          for (int n = 0; n < vmesh.GetSize(b, rad::prim::energy()); ++n) {
+            // Sync conserved and primitive density
```
pdmullen, line 315: Update comments and I think `_d` is confusing... can we switch to `_er` or something?

src/derived/fill_derived.cpp
```diff
@@ -163,6 +171,25 @@ void ConsToPrim(MeshData<Real> *md) {
             vel3 = vmesh(b, dust::cons::momentum(VI(n, 2)), k, j, i) / (w_d * hx[2]);
           }
         }
+        if (do_rad) {
+          for (int n = 0; n < vmesh.GetSize(b, rad::prim::energy()); ++n) {
+            // Set primitive density
+            const Real u_d = vmesh(b, rad::cons::energy(n), k, j, i);
```
pdmullen: same comment as below: 

Update comments and I think _d is confusing... can we switch to _er or something?

src/pgen/shock.hpp
```diff
@@ -74,18 +75,25 @@ inline void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
-  const bool do_radiation = artemis_pkg->Param<bool>("do_radiation");
+  const bool do_imc = artemis_pkg->Param<bool>("do_imc");
+  const bool do_moment = artemis_pkg->Param<bool>("do_moment");
   PARTHENON_REQUIRE(do_gas, "The shock problem requires gas hydrodynamics!");
   PARTHENON_REQUIRE(!(do_dust), "The shock problem does not permit dust hydrodynamics!");
   auto eos_d = pmb->packages.Get("gas")->Param<EOS>("eos_d");
 
+  Real ar = Null<Real>();
+  if (do_moment) {
+    ar = pmb->packages.Get("radiation")->Param<Real>("arad");
```
pdmullen, line 86: So this is not threaded through regression, right? I think that may not be a bad idea.  Will also test units with radiation.

src/artemis.cpp
```diff
@@ -70,6 +72,13 @@ Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
   const bool do_viscosity = pin->GetOrAddBoolean("physics", "viscosity", false);
   const bool do_conduction = pin->GetOrAddBoolean("physics", "conduction", false);
   const bool do_radiation = pin->GetOrAddBoolean("physics", "radiation", false);
+  // Check if we are using IMC or moment based radiation
+  // This maybe is not great if user doesn't comment out unused node...
+  const bool do_imc = do_radiation && pin->DoesBlockExist("jaybenne");
+  const bool do_moment = do_radiation && pin->DoesBlockExist("radiation");
```
pdmullen: How about temporarily naming it something more specific than radiation? e.g., `radiation_moments`?

src/radiation/moment/matter_coupling.hpp
```diff
@@ -0,0 +1,390 @@
...
+                              v0(b, gas::cons::momentum(2), k, j, i) / (hx[2] * dens)};
+        const Real Er0 = v0(b, rad::cons::energy(), k, j, i);
+        std::array<Real, 3> Fr0{v0(b, rad::cons::flux(0), k, j, i) / hx[0],
+                                v0(b, rad::cons::flux(1), k, j, i) / hx[1],
+                                v0(b, rad::cons::flux(2), k, j, i) / hx[2]};
+
+        // Note(AMD): There is some floating point difference between the internal energy
+        // used to compute the temperature and the internal energy obtained from that
+        // temperature: T = eos_d.TemperatureFromDensityInternalEnergy(dens, eg/dens); eg
+        // /= dens * eos_d.InternalEnergyFromDensityTemperature(dens,T)
```
pdmullen, line 107: this is unfortunate... any other construction to get around this?

pdmullen (commented):
First round of comments

src/artemis_driver.cpp
```diff
@@ -65,11 +66,19 @@ ArtemisDriver<GEOM>::ArtemisDriver(ParameterInput *pin, ApplicationInput *app_in
...
   do_diffusion = do_viscosity || do_conduction;
-  do_radiation = artemis_pkg->template Param<bool>("do_radiation");
-
+  do_imc = artemis_pkg->template Param<bool>("do_imc");
+  do_moment = artemis_pkg->template Param<bool>("do_moment");
+
+  if (do_moment) {
+    auto rad_int = pin->GetOrAddString("radiation", "integrator", "rk2");
+    PARTHENON_REQUIRE(((rad_int == "rk1") || (rad_int == "rk2") || (rad_int == "rk3")),
+                      "radiation/integrator must be rk1,rk2, or rk3.")
```
pdmullen, line 76: Are you purposefully disabling `rk4` and/or `vl2`?

src/artemis_driver.cpp
```diff
@@ -272,6 +284,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+
+  if (tm.ncycle % tm.ncycle_out == 0) {
+    if (Globals::my_rank == 0) {
+      std::cout << "\nTaking " << nsteps << " radiation substeps"
+                << " at dt=" << dtr << std::endl;
+    }
+  }
+
+  for (int step = 1; step <= nsteps; step++, trad += dtr) {
+    status = RadiationTasks().Execute();
```
pdmullen, line 320: I didn't read closely enough before... Are we confident that $\sum dt_\mathrm{step} = \Delta t$ ? Sometimes handling the last timestep in the subcycling can be tricky when dt_step ~ round off.

src/artemis_driver.cpp
```diff
@@ -272,6 +284,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+      auto &u0 = pmesh->mesh_data.GetOrAdd("u0", i);
+      auto &u1 = pmesh->mesh_data.GetOrAdd("u1", i);
+
+      // Start looking for incoming messages (including for flux correction)
+      auto start_recv = tl.AddTask(none, parthenon::StartReceiveBoundBufs<any>, u0);
+      auto start_flx_recv = tl.AddTask(none, parthenon::StartReceiveFluxCorrections, u0);
+
+      // Compute hydrodynamic fluxes
+      auto rad_flx = tl.AddTask(none, Radiation::CalculateFluxes, u0.get());
+      auto gas_flx = tl.AddTask(none, Gas::ZeroFluxes, u0.get());
```
pdmullen, line 368: Can you help me understand this?

src/utils/integrators/artemis_integrator.hpp
adamdempsey90: I had tried at one point to exclude variables here, but never got it to work. I agree this isn't ideal.

src/artemis_driver.cpp
adamdempsey90: I didn't know we supported rk4, do we?  These are the integrators that have an IMEX variant, so it excludes `vl2` for the radiation integrator (not the hydro integrator)

src/artemis_driver.cpp
adamdempsey90: Yes, I evenly divide the dt into N dt_step values above. So I'm not taking a small round off step at the end.

src/artemis_driver.cpp
adamdempsey90: The flux registers for `gas::cons::total_energy` and `gas::cons::momentum` are filled with the values from the hydro step. I need to zero those out here since the rad step only adds the RHS terms -- no fluxes for the gas.

src/artemis.cpp
adamdempsey90: See my other comment...I would vastly prefer `jaybenne` -> `radiation/imc` and `radiation` -> `radiation/moment`

src/derived/fill_derived.cpp
adamdempsey90: Sure, I can change. I kept the names to limit mistakes on my part

src/pgen/shock.hpp
adamdempsey90: Yeah, adding this to the tests is still on the todo

src/radiation/moment/radiation.cpp
adamdempsey90: It's a direct copy from density, so it should just work, right?

src/radiation/moment/radiation.cpp
adamdempsey90: It packs on different variables. I don't want to include dust variables for example.

src/radiation/moment/radiation.hpp
adamdempsey90: No, comment should be removed.

src/radiation/moment/radiation.cpp
adamdempsey90: This is unused now, I think, and can be removed

src/artemis_driver.cpp
pdmullen: It sounds like we should then have an `ApplySource` function that doesn't interface with the fluxes at all then... would that bypass the need to zero these fluxes?

src/radiation/moment/radiation.hpp
adamdempsey90: Does my outer loop over partitions not work? For cartesian, I just directly loop through all the meshblocks and compute the cell size without a kernel. And this is not a task, so the MPI call here should work, right?

src/utils/fluxes/riemann/llf.hpp
adamdempsey90: I'm not sure. Reusing the riemann solvers makes sense from the perspective that radiation is just a fluid in this scheme. All the variable names also make sense (density, velocity, pressure). The only slightly annoying part is that the rho*v*v part of the flux has a different factor on it than the pressure part, which is why these extra factors are needed.

I also don't want to duplicate the code.

src/radiation/moment/radiation.cpp
pdmullen: If you instead had a MeshData register containing only the subset of variables you'd like to update, I think you could simply pass it to our existing `ApplyUpdate` (as it exists in `develop`)

src/artemis.cpp
pdmullen: will merge that PR this week.

src/artemis_driver.cpp
pdmullen: I believe we do.  I remember I bugged you in the past on how to ensure the NBody integrator was compatible with RK4 coefficients.

src/radiation/moment/radiation.cpp
pdmullen: I have no doubt that they'd work, but I don't know if this is something we'd normally do.  

For most of our applications, I think we will be refining on density, pressure, (or maybe velocity), but radiation energy density?

src/derived/fill_derived.cpp
adamdempsey90: Done

src/utils/integrators/artemis_integrator.hpp
adamdempsey90: Do you know if there is an extra cost with multiple MeshData registers?

src/utils/fluxes/riemann/hlle.hpp
adamdempsey90: One option I can think of is after our reconstruction inner loop, we call another inner loop if the fluid is radiation and apply these constraints. Another alternative could be for that second inner loop to do PCM and swap based on some fluid specific heuristic .

src/artemis_driver.cpp
adamdempsey90: I need to update the radiation variables from the radiation fluxes, and that is done with `ApplyUpdate`, but my register contains the gas variables, so I need to zero the gas fluxes.

src/utils/fluxes/riemann/llf.hpp
adamdempsey90: I've split the riemann solvers up into different implementations for gas/dust and radiation fluids

src/utils/fluxes/riemann/hlle.hpp
adamdempsey90: fixed

src/artemis.cpp
adamdempsey90: Changed the names

src/pgen/shock.hpp
adamdempsey90: Tests added

src/utils/fluxes/riemann/hlle.hpp
adamdempsey90: done

src/utils/integrators/artemis_integrator.hpp
adamdempsey90: done

src/utils/fluxes/fluid_fluxes.hpp
adamdempsey90: This function is shared between all the fluids, so I'd prefer to keep this as is rather than introduce 4 new variables.

src/pgen/beam.hpp
adamdempsey90: Done

src/radiation/moment/matter_coupling.hpp
adamdempsey90: At the moment, I don't know. And I don't know if this is a problem on our side or in singularity. Maybe if in the ICs we did this, it would then always be consistent.

src/radiation/moment/radiation.cpp
adamdempsey90: I'm going to leave this as a future MR

src/utils/integrators/artemis_integrator.hpp
adamdempsey90: We can explore this in a future MR

src/utils/fluxes/fluid_fluxes.hpp
adamdempsey90: Leave for a future cleanup MR

adamdempsey90:
Not entirely sure what's going on with the CI. The new `radiation/rad_shock_cgs` test passes when I run it by hand. The darwin GPU CI is also not reporting for some reason, but I've run that whole `gpu.suite` by hand it it passes as well. 

So I think this is good to go in the sense that it has everything it needs and the tests all pass. There are certainly various improvements that can be done, but I will leave those for a future MR.

doc/src/physics.rst
```diff
@@ -843,6 +844,59 @@ Note that the grain sizes and density must be in code units.
...
+where the Eddington tensor is defined one of two ways:
+
+* P1: Isotropic pressure with :math:`\mathbf{f}= 1/3 \mathbf{I}` 
+* M1: Anisotropic pressure using the Levermore closure. 
+
+The parameter ``closure`` can be set to ``p1`` or ``m1`` to choose an appropriate closure model.
+
+In the moments approximation, the radiation is treated like a fluid and so has many of the typical fluid parameter options such as ``cfl``, `riemann``, ``reconstruct``, and ``nspecies``.
+
+If ``nspecies=1``, this tells |code| to use a gray description of the radiation.
```
pdmullen, line 879: Should we change it to `ngroups`? `nspecies` is more consistent with other "fluids".

doc/src/physics.rst
```diff
@@ -843,6 +844,59 @@ Note that the grain sizes and density must be in code units.
...
+   cfl = 0.3
+   riemann = hlle
+   reconstruct = plm
+   efloor = 1e-16
+   closure = p1
+   creduc = 100.0
+   full_coupling = false # simpler and faster matter-coupling
+
+
+Implict Monte Carlo
```
pdmullen, line 897: Implicit

src/artemis_driver.cpp
```diff
@@ -277,6 +290,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+    status = RadiationTasks().Execute();
+    if (status != TaskListStatus::complete) {
+      return status;
+    }
+  }
+  return status;
+}
+
+template <Coordinates GEOM>
+TaskCollection ArtemisDriver<GEOM>::RadiationTasks() {
```
pdmullen, line 329: The current `artemis` design has operator split `TaskCollection`s living in `.hpp` files (see, e.g., jaybenne and background linear advection).  Can you add this to `radiation/moments/moments.hpp` (or similar)?

src/artemis_driver.cpp
pdmullen: I'd like to switch to the `MeshData` formulation.  You are doing unnecessary work here and in flux correction, and `ApplyFluxDivergence` loses the abstraction we strived for... Can I make edits directly to this branch?

src/derived/fill_derived.cpp
```diff
@@ -163,6 +171,25 @@ void ConsToPrim(MeshData<Real> *md) {
             vel3 = vmesh(b, dust::cons::momentum(VI(n, 2)), k, j, i) / (w_d * hx[2]);
           }
         }
+        if (do_rad) {
+          for (int n = 0; n < vmesh.GetSize(b, rad::prim::energy()); ++n) {
+            // Set primitive density
```
pdmullen: // Set radiation energy density

src/artemis_driver.cpp
```diff
@@ -277,6 +290,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+          tl.AddTask(coupling, ArtemisDerived::SetAuxillaryFields<GEOM>, u0.get());
+
+      // Set (remaining) fields to be communicated
+      auto pre_comm = tl.AddTask(set_aux, PreCommFillDerived<MeshData<Real>>, u0.get());
+
+      // Set boundary conditions (both physical and logical)
+      auto bcs = parthenon::AddBoundaryExchangeTasks(pre_comm, tl, u0, pmesh->multilevel);
+
+      // Update primitive variables
+      auto c2p = tl.AddTask(TQ::local_sync, bcs, FillDerived<MeshData<Real>>, u0.get());
```
pdmullen: Don't need `TQ::local_sync`.

src/gas/gas.cpp
```diff
@@ -689,6 +690,23 @@ TaskStatus ZeroDiffusionFlux(MeshData<Real> *md) {
   return Diffusion::ZeroDiffusionImpl(md, vf);
 }
 
+//----------------------------------------------------------------------------------------
+//! \fn  TaskStatus Gas::ZeroDiffusionFlux
+//  \brief Resets the diffusion flux
+TaskStatus ZeroFluxes(MeshData<Real> *md) {
```
pdmullen, line 696: I want to eliminate this.

src/pgen/shock.hpp
pdmullen: The CI is complaining about `radiation/arad` not existing.  Is the failing test using moments? Because I guess `radiation/arad` is only stashed when invoking moments.

src/radiation/moment/matter_coupling.hpp
```diff
@@ -0,0 +1,417 @@
...
+
+          E += dE;
+          B += dB;
+          inner_err = std::max((std::abs(Fi) / etot), (c / chat * std::abs(Fr) / etot));
+          if (inner_err <= inner_tol) {
+            break;
+          }
+          //  printf("Continuing %d %d, %d %lg\n", j, i, inner_iter, inner_err);
+        }
+        //        printf("Inner done\n");
```
pdmullen: Can we cleanup old debug comments

src/radiation/moment/radiation.cpp
```diff
@@ -0,0 +1,455 @@
...
+  params.Add("rsolver", riemann_solver);
+
+  // Courant, Friedrichs, & Lewy (CFL) Number
+  const Real cfl_number = pin->GetOrAddReal("radiation/moment", "cfl", 0.8);
+  params.Add("cfl", cfl_number);
+
+  params.Add("full_coupling",
+             pin->GetOrAddBoolean("radiation/moment", "full_coupling", true));
+
+  const Real light = constants.GetCCode();
```
pdmullen, line 106: Can you write a detailed comment discussing units/constants? I am confused about the units stashed in params, the units that singularity has under the hood, and the constants invoked by singularity in the moments construction (e.g., c vs reduced speed of light)

src/radiation/moment/radiation.cpp
```diff
@@ -0,0 +1,455 @@
...
+  } else {
+    PARTHENON_FAIL("Riemann solver (radiation) not recognized.");
+  }
+  params.Add("rsolver", riemann_solver);
+
+  // Courant, Friedrichs, & Lewy (CFL) Number
+  const Real cfl_number = pin->GetOrAddReal("radiation/moment", "cfl", 0.8);
+  params.Add("cfl", cfl_number);
+
+  params.Add("full_coupling",
```
pdmullen, line 102: Can you add separation of params via comments.

src/radiation/moment/radiation.cpp
pdmullen: Remove.

src/radiation/moment/radiation.cpp
pdmullen: see above.

src/utils/fluxes/reconstruction/reconstruction.hpp
```diff
@@ -22,16 +22,70 @@ namespace ArtemisUtils {
...
+
+  if constexpr (is_grey<FLUID_TYPE>()) {
+    constexpr int nvar = 5;
+    const int nspecies = q.GetMaxNumberOfVars() / nvar;
+    for (int n = 0; n < nspecies; ++n) {
+      const int IER = n;
+      const int IFX = nspecies + (n * 3) + ((dir - 1));
+      const int IFY = nspecies + (n * 3) + ((dir - 1) + 1) % 3;
+      const int IFZ = nspecies + (n * 3) + ((dir - 1) + 2) % 3;
+      parthenon::par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, il, iu,
```
pdmullen, line 48: Do we need a separate kernel launch for this? Can we handle this inside the normal reconstruction under a constexpr for rad?

pdmullen (commented):
A lot here...

doc/src/physics.rst
adamdempsey90: I'd prefer to keep it nspecies to be consistent with the other fluid types

doc/src/physics.rst
adamdempsey90: fixed

src/artemis_driver.cpp
adamdempsey90: I prefer all the driver tasks be in one place, not scattered about. I didn't catch that with the advection step, but I should have.

src/artemis_driver.cpp
adamdempsey90: If you want to make those changes, then can you open another MR that points to this branch?

src/artemis_driver.cpp
adamdempsey90: removed

src/derived/fill_derived.cpp
adamdempsey90: fixed

src/radiation/moment/matter_coupling.hpp
adamdempsey90: done

src/radiation/moment/radiation.cpp
adamdempsey90: The units I stash in params are there for analysis, specifically c and arad. I want the code values for those numbers so that I can convert the code outputs of the primitives.

src/radiation/moment/radiation.cpp
adamdempsey90: done

src/radiation/moment/radiation.cpp
adamdempsey90: I put in a comment, too

src/radiation/moment/radiation.cpp
adamdempsey90: I'm going to keep this since there are test problems out there that refine on energy density. I see no harm in keeping them.

src/utils/fluxes/reconstruction/reconstruction.hpp
adamdempsey90: I think it's better separate since I'd have to put that in all three reconstruction kernels, instead of of just one. This also demonstrates a way to fall back to PCM for other fluids like dust.

pdmullen: It's an extra kernel launch and barrier that I'd love to avoid.

src/artemis_driver.cpp
pdmullen: I have to disagree here.  

There is a nice separation now, where `artemis_driver.cpp` only contains the main driver plus its calls to `.Execute()` for operator split task collections.  We wouldn't want, for example, all of the `jaybenne` task collection inside of `artemis_driver.cpp`.  

For this MR, I'd like to keep things consistent with the current design (rather than bloating `artemis_driver.cpp` with our other `TaskCollections`).  I can do this instead of you in my MR into your branch, if you'd like.

src/artemis_driver.cpp
```diff
@@ -277,6 +290,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+    for (int i = 0; i < num_partitions; i++) {
+      auto &tl = tr[i];
+      auto &u0 = pmesh->mesh_data.GetOrAdd("u0", i);
+      auto &u1 = pmesh->mesh_data.GetOrAdd("u1", i);
+
+      // Start looking for incoming messages (including for flux correction)
+      auto start_recv = tl.AddTask(none, parthenon::StartReceiveBoundBufs<any>, u0);
+      auto start_flx_recv = tl.AddTask(none, parthenon::StartReceiveFluxCorrections, u0);
+
+      // Compute hydrodynamic fluxes
```
RyanWollaeger, line 366:
```suggestion
      // Compute radiation fluxes
```
?

src/artemis_driver.cpp
```diff
@@ -277,6 +290,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+      // Apply flux divergence
+      auto update = tl.AddTask(rad_flx | gas_flx | set_flx, Radiation::ApplyUpdate<GEOM>,
+                               u0.get(), u1.get(), stage, gam0, gam1, bdt);
+
+      // Apply "coordinate source terms"
+      auto coord_src = tl.AddTask(update, Radiation::FluxSource, u0.get(), bdt);
+
+      // Apply matter-coupling step
+      // This is the first task to update the gas/dust values
+      auto coupling =
```
RyanWollaeger, line 386: Assuming reading this correctly, here is a 0D iterative coupling step - does this mean the radiation solver can super-cycle hydro?

src/artemis_driver.cpp
```diff
@@ -277,6 +290,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+  // Deep copy u0 into u1 for integrator logic
+  auto &init_region = tc.AddRegion(num_partitions);
+  for (int i = 0; i < num_partitions; i++) {
+    auto &tl = init_region[i];
+    auto &u0 = pmesh->mesh_data.GetOrAdd("u0", i);
+    auto &u1 = pmesh->mesh_data.GetOrAdd("u1", i);
+    tl.AddTask(none, ArtemisUtils::DeepCopyConservedData, u1.get(), u0.get());
+  }
+
+  // Now do explicit integration of unsplit physics
```
RyanWollaeger, line 349:
```suggestion
  // Now do IMEX integration of coupled radiation equations
```
?

src/artemis_driver.cpp
```diff
@@ -277,6 +290,120 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
...
+      auto start_flx_recv = tl.AddTask(none, parthenon::StartReceiveFluxCorrections, u0);
+
+      // Compute hydrodynamic fluxes
+      auto rad_flx = tl.AddTask(none, Radiation::CalculateFluxes, u0.get());
+      auto gas_flx = tl.AddTask(none, Gas::ZeroFluxes, u0.get());
+
+      // Communicate and set fluxes
+      auto send_flx =
+          tl.AddTask(rad_flx | gas_flx,
+                     parthenon::SendBoundBufs<parthenon::BoundaryType::flxcor_send>, u0);
```
RyanWollaeger, line 373: Still Parthenon-illiterate: does this communcate gas fluxes as well, or just rad fluxes?

src/artemis_driver.cpp
pdmullen: Addressed in #74

src/utils/fluxes/reconstruction/reconstruction.hpp
pdmullen: Cleaned up sorta in #74

src/artemis_driver.cpp
pdmullen: Addressed in #74

src/artemis_driver.cpp
pdmullen: The radiation source term is locally implicit (as opposed to globally implicit).  I am a bit unsure of terminology super-cycling (maybe we can chat offline?)... but I am confident that the current moments implementation sub-cycles the unsplit timestep.

src/artemis_driver.cpp
pdmullen: Addressed in #74

src/artemis_driver.cpp
pdmullen: It communicates all fluxes as written.  This is addressed in #74

src/gas/gas.cpp
pdmullen: Eliminated in #74.

pdmullen (approved):
Approving now.  Let's bring in #74 shortly thereafter.

PR #59, opened 2025-03-06
[BugFix] Fix units of G and make strat pgen work in cgs

adamdempsey90:
## Background

We had the wrong units for the gravitational constant. This fixes that, puts in the modifications needed for the strat pgen to work in cgs, and does some other housekeeping.

The disk tests now are run in ppd cgs units.

## Checklist

### DIFF
--- a/inputs/disk/disk_axi.in
+++ b/inputs/disk/disk_axi.in
@@ -14,6 +14,8 @@
 <artemis>
 problem = disk              # name of the pgen
 coordinates = axisymmetric  # coordinate system
+physical_units = cgs
+unit_conversion = ppd
 
 <parthenon/job>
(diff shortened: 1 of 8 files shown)


pdmullen (approved):
LGTM! I was surprised to see the removal of `KOKKOS_FUNCTION`s from some ctors.  Can you advise on why these are unnecessary?

adamdempsey90:
> LGTM! I was surprised to see the removal of `KOKKOS_FUNCTION`s from some ctors. Can you advise on why these are unnecessary?

This is what the darwin volta compiler said,
```
warning: __device__ annotation is ignored on a function("Units") that is explicitly defaulted on its first declaration
warning: __host__ annotation is ignored on a function("Units") that is explicitly defaulted on its first declaration
```

PR #60, opened 2025-03-07
More portable restarts

adamdempsey90:
## Background

This changes the rebound restart file in params from a `std::vector<char>` to a `std::vector<uint8_t>` which should be more portable. Because this would break existing restart files, I added an option to not do this. 

This updates parthenon to the latest develop

## Checklist


adamdempsey90:
Looks like parthenon is not outputting the reb sim buffer when it is a vector of uint8_t or unsigned char. It doesn't complain about anything, it just doesn't write it into the restart file.

adamdempsey90:
> Looks like parthenon is not outputting the reb sim buffer when it is a vector of uint8_t or unsigned char. It doesn't complain about anything, it just doesn't write it into the restart file.

Yeah, parthenon definitely doesn't support `uint8_t` or `unsigned char` in its hdf5 output. We'll need to add that if we want to do this. @Yurlungur, thoughts / objections?

adamdempsey90:
@pdmullen Ready to go

adamdempsey90:
> @pdmullen Ready to go

Ah, I guess we need those other changes for the new parthenon that you're working on

adamdempsey90:
@pdmullen can you approve and merge

pdmullen (approved):

PR #61, opened 2025-03-24
Update to use mean opacity variants

pdmullen:
## Background
Previously, `artemis` and `jaybenne` somewhat "cheated" a gray formulation of IMC by interfacing with mutifrequency opacity objects and just passing a `null` frequency argument to them.  We could get away with this because all of our previous use cases used constant opacities (independent of frequency) and/or powerlaw opacities (in density and temperature).  

This was, however, not a long term solution... and `jaybenne` has since then starting moving towards unique gray vs multigroup representations.  On this trajectory, `singularity-opac` now provides a `MeanOpacity` variant which constructs a table from an underlying opacity model.  For example, given a constant specific absorption opacity $\kappa_a$,  the `MeanOpacity` variant produces Planck and Rosseland means of the constant opacity, producing a table of constant values.  Albeit this overkill for this use case, the infrastructure permits us to interface with more complex opacity data and move towards our production calculations.

## Description of Changes
Gone are the null frequency arguments to opacity objects for gray IMC.  We now interface with `MeanOpacity` objects that are only supplied density and temperature arguments.  Note that this means that every gray opacity we work with is now tabular. 

Also, along the way, @RyanWollaeger and I cleaned up sourcing in `jaybenne`.    Before, `jaybenne` would always source the same number of particles per timestep.  Now we more carefully try to maintain the `jaybenne/num_particles` target.  This is not only via sourcing, but also new population control routines in `jaybenne`.  You'll notice that I have incremented the number of particles in some tests in this MR.  This is because the tests were using many more particles than we previously thought in previous implementations, associated with the erroneous sourcing behavior described above. 

I worry there are some knobs that we have introduced in the path towards this MR (and many of the knobs we have decided to not yet be user-interfaceable)... but the path to deciding how to best interface with this new utility is in application.  

## Checklist

### DIFF
--- a/src/gas/gas.cpp
+++ b/src/gas/gas.cpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2023-2024. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2023-2025. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/src/gas/gas.cpp
+++ b/src/gas/gas.cpp
@@ -129,46 +129,69 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
     params.Add("adiabatic_index", gamma);
   }
 
-  // Absorption opacity model
-  // TODO(@pdmullen): This may not be the right place for this... how about dust opacity?
-  ArtemisUtils::Opacity opacity;
-  std::string opacity_model_name =
-      pin->GetOrAddString("gas/opacity/absorption", "opacity_model", "constant");
-  const Real length = units.GetLengthCodeToPhysical();
+  // Opacity models
   const Real time = units.GetTimeCodeToPhysical();
   const Real mass = units.GetMassCodeToPhysical();
+  const Real length = units.GetLengthCodeToPhysical();
   const Real temp = units.GetTemperatureCodeToPhysical();
+
+  // Absorption opacity model
+  ArtemisUtils::Opacity model;
+  std::string opacity_model_name =
+      pin->GetOrAddString("gas/opacity/absorption", "opacity_model", "constant");
   if (opacity_model_name == "none") {
-    opacity = NonCGSUnits<Gray>(Gray(0.0), time, mass, length, temp);
+    model = Gray(0.0);
   } else if (opacity_model_name == "constant") {
     const Real kappa_a = pin->GetOrAddReal("gas/opacity/absorption", "kappa_a", 0.0);
-    opacity = NonCGSUnits<Gray>(Gray(kappa_a), time, mass, length, temp);
+    model = Gray(kappa_a);
   } else if (opacity_model_name == "powerlaw") {
     const Real coef_kappa_a =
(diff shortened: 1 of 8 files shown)


src/gas/gas.cpp
```diff
@@ -129,46 +129,69 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
...
   } else {
     PARTHENON_FAIL("Opacity model not recognized!");
   }
+  // Instantiate mean absorption opacity object (i.e., table)
+  const Real lRhoMin_a = pin->GetOrAddReal("gas/opacity/absorption", "lRhoMin", -1.0);
+  const Real lRhoMax_a = pin->GetOrAddReal("gas/opacity/absorption", "lRhoMax", 1.0);
+  const int NRho_a = pin->GetOrAddInteger("gas/opacity/absorption", "NRho", 2);
+  const Real lTMin_a = pin->GetOrAddReal("gas/opacity/absorption", "lTMin", -1.0);
+  const Real lTMax_a = pin->GetOrAddReal("gas/opacity/absorption", "lTMax", 1.0);
+  const int NT_a = pin->GetOrAddInteger("gas/opacity/absorption", "NT", 2);
```
adamdempsey90, line 162: in the situation of an actual tabular opacity, can this information just be stored in the header of the file? It would be nice to not have to specify these things in the input file

tst/scripts/radiation/thermalization.py
```diff
@@ -78,7 +78,7 @@ def run(**kwargs):
         "gas/opacity/absorption/opacity_model=constant",
         "gas/opacity/absorption/kappa_a={:24.16e}".format(_ka),
         "jaybenne/dt=1.0e-10",
-        "jaybenne/num_particles=1000",
+        "jaybenne/num_particles=200000",
```
adamdempsey90, line 81: Is this not too big for the CI?

pdmullen: The current version of this test in `develop` approaches this total number of particles due to the sourcing issues discussed in the MR body.

src/gas/gas.cpp
pdmullen: Good idea.  I'd like to loop in @RyanWollaeger here and get his opinion as well.  

My thoughts: the underlying opacity data likely has some resolution structure (and it may be conglomerated from multiple sources; c.f., the PPD opacity tables).  We likely cannot adjust this root data's tabular size (i.e., it is just whatever is provided or available to us).  What we do have the freedom of is in determining the size of the table that is generated on-the-fly at the beginning of an `artemis` simulation from this root data... that is what is done by specifying the highlighted parameters above.  To your point, if we already have in existence some table that is already in our format and we just want to read it in, then I think your design idea is optimal.  But then multigroup confuses things... e.g., if we want to change our group structure on the fly, then pre-existing tables may not be the right path.  This conversation approaches one that we might need to have with our opacity team as well.

RyanWollaeger: Supplying this table bound/size data should only correspond to one constructor overload of the mean opacity class, whereas the other takes a file name and gets this data from that file (if recalled correctly). The file is either hdf5 or ascii, and I think the numerical sizes are in the header in ascii (and encoded in the hdf5 similarly via DataBox).

adamdempsey90 (approved):

PR #64, opened 2025-03-30
[BugFix] Viscosity

adamdempsey90:
## Background

The units were wrong on the strain tensor when there were non-unity scale factors. This was because we were using the cartesian distance to get the `dx` in `dv/dx`. This switches to using coordinate differences. 

A small reproducer seems to be behaving better with this change. 

## Checklist


pdmullen (approved):
LGTM! Was this actually a units issue? It looks more like the wrong delta was previously used? Not important, just want to make sure I understand.

adamdempsey90:
> LGTM! Was this actually a units issue? It looks more like the wrong delta was previously used? Not important, just want to make sure I understand.

Overloaded term. In this instance, by units I meant there were too many length factors  for the strain tensor when the scale factor was r and dx was the real distance.

PR #65, opened 2025-04-18
Aandrews/frontier build

amaya-the-grey:
## Checklist

With `frontier-gpu-release` presets, 18/18 tests pass. However, `frontier-gpu-debug` has issues and can't quite run the tests. In the interest of time here is what I have so far to get you all up and running on frontier! Not sure what capacity debug builds and tests hold in your workflow but I will keep working on it.

Please let me know your thoughts or if you would like to see any changes.

Thanks! ^.^

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -131,7 +131,17 @@
         "CMAKE_MAKE_PROGRAM": "$env{MAKE_PROGRAM}",
         "CMAKE_BUILD_TYPE": "Release",
         "NUM_GPU_DEVICES_PER_NODE": "4",
-        "CMAKE_INSTALL_PREFIX": "/lustre/orion/ast222/proj-shared/artemis/dependencies/"
+        "CMAKE_INSTALL_PREFIX": "/lustre/orion/ast222/proj-shared/jovian/dependencies/"
+      }
+    },
+    {
+      "name": "frontier-gpu-debug",
+      "cacheVariables": {
+        "CMAKE_MAKE_PROGRAM": "$env{MAKE_PROGRAM}",
+        "CMAKE_BUILD_TYPE": "Debug",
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -124,6 +124,25 @@
...
+        "CMAKE_INSTALL_PREFIX": "/lustre/orion/ast222/proj-shared/jovian/dependencies/"
+      }
+    },
+    {
+      "name": "frontier-gpu-debug",
+      "cacheVariables": {
+        "CMAKE_MAKE_PROGRAM": "$env{MAKE_PROGRAM}",
+        "CMAKE_BUILD_TYPE": "Debug",
+        "NUM_GPU_DEVICES_PER_NODE": "4",
+        "CMAKE_INSTALL_PREFIX": "/lustre/orion/ast222/proj-shared/jovian/dependencies/",
```
adamdempsey90: Let's make the path to this directory an environment variable that's defined in the user's bashrc. That will make it easy to swap out for different directories (like a different project number). Maybe call it `"$env{JOVIAN_DIR}/dependencies"'.  

Similar change in the `env/bash` script.

amaya-the-grey: Sure! What should the environment variable be named? Something like `JOVIAN_DEP_PATH`?

adamdempsey90: Yeah, that's fine. And actually, do we need to set `CMAKE_INSTALL_PREFIX` at all? We don't usually do `make install`. Did you need this set to get it to build?

amaya-the-grey: Ah, you can build just fine without it, but you do need to specify it to install. I was just following the workflow in the README but if installation is not necessary then we can cut it. Up to you all! ^.^

adamdempsey90 (approved):

PR #66, opened 2025-04-20
Improvements to shearing box and new pgens

adamdempsey90:
## Background

We were seeing some vertical asymmetries develop in the shearing box. I think those were largely due to our extrapolation BCs. This PR improves all of the BCs used in the shearing box and the asymmetries are under better control. The biggest conceptual change is that at the vertical boundaries the temperature is now fixed to the initial value rather than using the current temperature. This was the biggest stabilizer. 

This PR also includes the Liska-Wendroff, Rayleigh-Taylor, and Kelvin-Helmoltz test problems. All show good symmetry preservation. 

## Checklist

### DIFF
--- a/src/rotating_frame/rotating_frame_impl.hpp
+++ b/src/rotating_frame/rotating_frame_impl.hpp
@@ -53,60 +53,51 @@ TaskStatus ShearingBoxImpl(MeshData<Real> *md, const Real om0, const Real qshear
         // Extract coordinates
         geometry::Coords<Coordinates::cartesian> coords(vmesh.GetCoordinates(b), k, j, i);
 
-         const Real dx = coords.bnds.x1[1] - coords.bnds.x1[0];
-         const Real dz = coords.bnds.x3[1] - coords.bnds.x3[0];
+        const Real dx = coords.bnds.x1[1] - coords.bnds.x1[0];
+        const Real dz = coords.bnds.x3[1] - coords.bnds.x3[0];
 
-         const Real xc = 0.5 * (coords.bnds.x1[1] + coords.bnds.x1[0]);
-         const Real zc = 0.5 * (coords.bnds.x3[1] + coords.bnds.x3[0]);
+        const Real xc = 0.5 * (coords.bnds.x1[1] + coords.bnds.x1[0]);
+        const Real zc = 0.5 * (coords.bnds.x3[1] + coords.bnds.x3[0]);
 
-         const Real phi_xm1 = -qshear * omsq * SQR(coords.bnds.x1[0]);
-         const Real phi_xc = -qshear * omsq * SQR(xc);
-         const Real phi_xp1 = -qshear * omsq * SQR(coords.bnds.x1[1]);
+        const Real phi_xm1 = -qshear * omsq * SQR(coords.bnds.x1[0]);
+        const Real phi_xc = -qshear * omsq * SQR(xc);
+        const Real phi_xp1 = -qshear * omsq * SQR(coords.bnds.x1[1]);
 
-         const Real phi_zm1 = 0.5 * omsq * SQR(coords.bnds.x3[0]);
-         const Real phi_zc = 0.5 * omsq * SQR(zc);
-         const Real phi_zp1 = 0.5 * omsq * SQR(coords.bnds.x3[1]);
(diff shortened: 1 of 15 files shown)


src/rotating_frame/rotating_frame_impl.hpp
```diff
@@ -50,26 +55,51 @@ TaskStatus ShearingBoxImpl(MeshData<Real> *md, const Real om0, const Real qshear
...
+        const Real phi_xm1 = -qshear * omsq * SQR(coords.bnds.x1[0]);
+        const Real phi_xc = -qshear * omsq * SQR(xc);
+        const Real phi_xp1 = -qshear * omsq * SQR(coords.bnds.x1[1]);
+
+        const Real phi_zm1 = 0.5 * omsq * SQR(coords.bnds.x3[0]);
+        const Real phi_zc = 0.5 * omsq * SQR(zc);
+        const Real phi_zp1 = 0.5 * omsq * SQR(coords.bnds.x3[1]);
+
+        const Real dpxc = (phi_xp1 - phi_xm1) / dx;
+        const Real dpxm = (phi_xc - phi_xm1) / dx;
```
pdmullen, line 71: I don't follow this one... shouldn't this be centered on the cell face?

`(phi_xc - phi_xcm1) / dx`

?

src/utils/artemis_utils.cpp
```diff
@@ -62,7 +62,7 @@ void PrintArtemisConfiguration(Packages_t &packages) {
       auto particles = nbody_pkg->Param<ParArray1D<NBody::Particle>>("particles");
       auto particles_h = particles.GetHostMirrorAndCopy();
       auto npart = particles_h.size();
-      printf("      %d NBody particle(s)\n", npart);
+      printf("      %d NBody particle(s)\n", static_cast<int>(npart));
```
pdmullen, line 65: thank you

src/rotating_frame/rotating_frame_impl.hpp
adamdempsey90: This is from athena https://github.com/PrincetonUniversity/Athena-Cversion/blob/master/src/integrators/integrate_2d_vl.c#L479 . It's supposed to be the energy conserving way of doing this

pdmullen: You are right.  It is Equation (36) from here: https://iopscience.iop.org/article/10.1088/0067-0049/189/1/142

I guess my brain was too locked into self-gravity, where I usually define...

`S_E = 1/2 [ (rho vx)_{i+1/2} gx_{i+1/2} + (rho vx)_{i-1/2} gx_{i-1/2} ]`

where `gx_{i+1/2}` is defined at the cell face `-(phi_{i+1} + phi_{i}) / dx`

One can recast this as 

`S_E = -1/2 [(rho vx)_{i+1/2} (phi_r - phi_{i}) +  (rho vx)_{i-1/2} (phi_{i} - phi_l)]`

where `phir = 1/2 (phi_{i+1} + phi_{i})`.  

But since shearing box has an analytic solution for the potential (as opposed to in self-gravity, the potential is defined on cell centers), it appears that athena takes phir = phi(x_c + 0.5dx)... and as the paper above argues, this is conservative...

Just a bit of a mind-bend, as at first glance it feels like the centering of the finite difference is off...

src/rotating_frame/rotating_frame_impl.hpp
```diff
@@ -50,26 +55,51 @@ TaskStatus ShearingBoxImpl(MeshData<Real> *md, const Real om0, const Real qshear
...
+                rdt * (dpxc - 2.0 * om0 * v2);
             vmesh(b, gas::cons::momentum(VI(n, 1)), k, j, i) -= rdt * 2.0 * om0 * v1;
-            vmesh(b, gas::cons::momentum(VI(n, 2)), k, j, i) -= rdt * dpz;
+            vmesh(b, gas::cons::momentum(VI(n, 2)), k, j, i) -= rdt * dpzc;
+
             vmesh(b, gas::cons::total_energy(n), k, j, i) -=
-                rdt * (v1 * dpx + v3 * dpz); // NOTE(ADM): Change to use the fluxes
+                rdt * (v1 * dpxc + v3 * dpzc);
+
+            // TODO: Why does this not work
```
pdmullen, line 93: Can we eliminate this until it is working?

pdmullen (approved):

PR #68, opened 2025-04-29
Algorithmic Changes for Shearing Box

pdmullen:
## Checklist

### DIFF
--- a/src/artemis.hpp
+++ b/src/artemis.hpp
@@ -92,6 +92,8 @@ enum class Coordinates {
 };
 // ...Riemann solvers
 enum class RSolver { hllc, hlle, llf, null };
+// ... Upwinding
+enum class Upwind { l, r };
 // ...Reconstrution algorithms
 enum class ReconstructionMethod { pcm, plm, ppm, null };
 // ...Fluid types
--- a/src/artemis.hpp
+++ b/src/artemis.hpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2023-2024. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2023-2025. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
@@ -92,8 +92,8 @@ enum class Coordinates {
 };
 // ...Riemann solvers
 enum class RSolver { hllc, hlle, llf, null };
-// ... Upwinding
-enum class Upwind { l, r };
+// ... Upwinding (left vs right state)
+enum class Upwind { l, r, null };
 // ...Reconstrution algorithms
 enum class ReconstructionMethod { pcm, plm, ppm, null };
 // ...Fluid types
--- a/src/dust/dust.cpp
+++ b/src/dust/dust.cpp
@@ -278,7 +278,8 @@ Real EstimateTimestepMesh(MeshData<Real> *md) {
           Real denom = 0.0;
           for (int d = 0; d < ndim; d++) {
             Real vd = vmesh(b, dust::prim::velocity(VI(n, d)), k, j, i);
-            vd += (do_shear * (d == 1)) * ShearVelocity<GEOM>(qshear, om0, coords.x1v());
+            vd += (do_shear && (d == 1)) ? ShearVelocity<GEOM>(qshear, om0, coords.x1v())
+                                         : 0.0;
             denom += std::abs(vd) / dx[d];
           }
           ldt = std::min(ldt, 1.0 / denom);
--- a/src/dust/dust.cpp
+++ b/src/dust/dust.cpp
@@ -239,7 +239,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
 template <Coordinates GEOM>
 Real EstimateTimestepMesh(MeshData<Real> *md) {
   using parthenon::MakePackDescriptor;
-  using RotatingFrame::ShearVelocity;
+  using RotatingFrame::BackgroundVelocity;
   auto pm = md->GetParentPointer();
   auto &resolved_pkgs = pm->resolved_packages;
 
@@ -283,8 +283,8 @@ Real EstimateTimestepMesh(MeshData<Real> *md) {
         }
 
         if (do_shear) {
-          ldt = std::min(
-              ldt, dx[1] / std::abs(ShearVelocity<GEOM>(qshear, om0, coords.x1v())));
+          const auto ww = BackgroundVelocity<GEOM>(qshear, om0, coords.x1v());
+          ldt = std::min(ldt, dx[1] / std::abs(ww[1]));
         }
       },
       Kokkos::Min<Real>(min_dt));
--- a/src/dust/dust.cpp
+++ b/src/dust/dust.cpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2023-2024. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2023-2025. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
@@ -22,6 +22,7 @@
 #include "artemis.hpp"
 #include "dust/dust.hpp"
 #include "geometry/geometry.hpp"
+#include "rotating_frame/rotating_frame.hpp"
 #include "utils/artemis_utils.hpp"
 #include "utils/fluxes/fluid_fluxes.hpp"
 #include "utils/history.hpp"
@@ -238,6 +239,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
 template <Coordinates GEOM>
 Real EstimateTimestepMesh(MeshData<Real> *md) {
   using parthenon::MakePackDescriptor;
+  using RotatingFrame::ShearVelocity;
   auto pm = md->GetParentPointer();
   auto &resolved_pkgs = pm->resolved_packages;
 
@@ -247,13 +249,12 @@ Real EstimateTimestepMesh(MeshData<Real> *md) {
   // NOTE(@pdmullen): Without FARGO, integration of the residual eqs must have a dt
   // limited by vy = dvy + vy0 = dvy - q Omega x
   bool do_shear = false;
-  Real qomega = 0.0;
+  Real qshear = 0.0, om0 = 0.0;
   if (pm->packages.Get("artemis")->Param<bool>("do_rotating_frame")) {
     auto &rframe_pkg = pm->packages.Get("rotating_frame");
-    const Real omega = rframe_pkg->Param<Real>("omega");
-    const Real qshear = rframe_pkg->Param<Real>("qshear");
-    qomega = qshear * omega;
-    do_shear = (qomega >= 0.0);
+    qshear = rframe_pkg->Param<Real>("qshear");
+    om0 = rframe_pkg->Param<Real>("omega");
+    do_shear = (qshear * om0 >= 0.0);
   }
 
   static auto desc =
@@ -277,7 +278,7 @@ Real EstimateTimestepMesh(MeshData<Real> *md) {
           Real denom = 0.0;
           for (int d = 0; d < ndim; d++) {
             Real vd = vmesh(b, dust::prim::velocity(VI(n, d)), k, j, i);
-            vd -= (do_shear) * (qomega * coords.x1v());
+            vd += (do_shear * (d == 1)) * ShearVelocity<GEOM>(qshear, om0, coords.x1v());
             denom += std::abs(vd) / dx[d];
           }
           ldt = std::min(ldt, 1.0 / denom);
--- a/src/dust/dust.cpp
+++ b/src/dust/dust.cpp
@@ -254,7 +254,7 @@ Real EstimateTimestepMesh(MeshData<Real> *md) {
     auto &rframe_pkg = pm->packages.Get("rotating_frame");
     qshear = rframe_pkg->Param<Real>("qshear");
     om0 = rframe_pkg->Param<Real>("omega");
-    do_shear = (qshear * om0 >= 0.0);
+    do_shear = (qshear != 0.0);
   }
 
   static auto desc =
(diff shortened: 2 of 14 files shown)


src/dust/dust.cpp
```diff
@@ -264,7 +276,9 @@ Real EstimateTimestepMesh(MeshData<Real> *md) {
         for (int n = 0; n < vmesh.GetSize(b, dust::prim::density()); ++n) {
           Real denom = 0.0;
           for (int d = 0; d < ndim; d++) {
-            denom += std::abs(vmesh(b, dust::prim::velocity(VI(n, d)), k, j, i)) / dx[d];
+            Real vd = vmesh(b, dust::prim::velocity(VI(n, d)), k, j, i);
+            vd -= (do_shear) * (qomega * coords.x1v());
```
adamdempsey90: Can we make a utility function in `rotating_frame` that calculates the background velocity? It should be templated on geom and passed omega and q and x.

src/artemis.hpp
```diff
@@ -92,6 +92,8 @@ enum class Coordinates {
 };
 // ...Riemann solvers
 enum class RSolver { hllc, hlle, llf, null };
+// ... Upwinding
+enum class Upwind { l, r };
```
adamdempsey90:
```suggestion
enum class Upwind { left, right };
```

src/rotating_frame/rotating_frame.hpp
```diff
@@ -14,38 +14,282 @@
...
+    const Real &u1_et = v1(b, gas::cons::total_energy(n), k, j, i);
+    const Real &u1_ei = v1(b, gas::cons::internal_energy(n), k, j, i);
+
+    u0_dn = g0 * u0_dn + g1 * u1_dn + fac * (qp[0] - q[0]);
+    u0_m1 = g0 * u0_m1 + g1 * u1_m1 + fac * (qp[1] - q[1]);
+    u0_m2 = g0 * u0_m2 + g1 * u1_m2 + fac * (qp[2] - q[2]);
+    u0_m3 = g0 * u0_m3 + g1 * u1_m3 + fac * (qp[3] - q[3]);
+    u0_et = g0 * u0_et + g1 * u1_et + fac * (qp[4] - q[4]);
+    u0_ei = g0 * u0_ei + g1 * u1_ei + fac * (qp[5] - q[5]);
+  } else {
```
adamdempsey90:
```suggestion
  } else if constexpr (FLUID_TYPE == Fluid::dust) {
```

src/rotating_frame/rotating_frame.hpp
```diff
@@ -14,38 +14,282 @@
...
+    u0_m3 = g0 * u0_m3 + g1 * u1_m3 + fac * (qp[3] - q[3]);
+  }
+}
+
+//----------------------------------------------------------------------------------------
+//! \fn  ReconSingle
+//! \brief Invoke PLM reconstruction for linear advection
+//! NOTE(@pdmullen): Donor cell reconstruction *could* nestle into the stencil of the
+//! Upwind function, but higher order reconstruction (e.g., PPM) is incompatible... Do we
+//! want to support all reconstruction options?
```
adamdempsey90, line 124: Typically this part is done with PPM

src/rotating_frame/rotating_frame.hpp
```diff
@@ -14,38 +14,282 @@
...
+                                              const int n, const int k, const int j,
+                                              const int i) {
+  if constexpr (FLUID_TYPE == Fluid::gas) {
+    ReconSingle<UDIR, gas::cons::density>(vmesh, arr, 0, b, n, k, j, i);
+    ReconSingle<UDIR, gas::cons::momentum>(vmesh, arr, 1, b, VI(n, 0), k, j, i);
+    ReconSingle<UDIR, gas::cons::momentum>(vmesh, arr, 2, b, VI(n, 1), k, j, i);
+    ReconSingle<UDIR, gas::cons::momentum>(vmesh, arr, 3, b, VI(n, 2), k, j, i);
+    ReconSingle<UDIR, gas::cons::total_energy>(vmesh, arr, 4, b, n, k, j, i);
+    ReconSingle<UDIR, gas::cons::internal_energy>(vmesh, arr, 5, b, n, k, j, i);
+  } else {
```
adamdempsey90:
```suggestion
  } else if constexpr (FLUID_TYPE == Fluid::dust) {
```

src/rotating_frame/rotating_frame.hpp
```diff
@@ -14,38 +14,282 @@
...
+
+//----------------------------------------------------------------------------------------
+//! \fn  Upwind
+//! \brief
+template <Fluid FLUID_TYPE, Upwind UDIR, typename V1>
+KOKKOS_INLINE_FUNCTION void Upwind(const V1 &v0, const V1 &v1, const Real g0,
+                                   const Real g1, const Real wdt, const int b,
+                                   const int k, IndexRange jb, const int i) {
+  // fluid indexing
+  int nu = Null<int>();
```
adamdempsey90, line 165: interesting variable name choice ;)

src/rotating_frame/rotating_frame.hpp
```diff
@@ -14,38 +14,282 @@
...
+  const int joff = r_stencil - l_stencil;
+
+  // reconstruct and advance
+  for (int n = 0; n < nu; ++n) {
+    auto uup = NewArray<Real, 6>();
+    UpwindReconstruct<FLUID_TYPE, UDIR>(v0, uup, b, n, k, jstart + joff, i);
+    auto uu = NewArray<Real, 6>();
+    UpwindReconstruct<FLUID_TYPE, UDIR>(v0, uu, b, n, k, jstart, i);
+    auto uum = NewArray<Real, 6>();
+    for (int j = jb.s; j < jb.e; ++j) {
```
adamdempsey90, line 184: just checking this isn't `j <= jb.e`?

src/rotating_frame/rotating_frame_impl.hpp
```diff
@@ -24,104 +24,79 @@
...
-
-        const Real dpxc = (phi_xp1 - phi_xm1) / dx;
-        const Real dpxm = (phi_xc - phi_xm1) / dx;
-        const Real dpxp = (phi_xp1 - phi_xc) / dx;
-
-        const Real dpzc = three_d * (phi_zp1 - phi_zm1) / dz;
-        const Real dpzm = three_d * (phi_zc - phi_zm1) / dz;
-        const Real dpzp = three_d * (phi_zp1 - phi_zc) / dz;
+        // Evaluate vertical gravity (at cell center)
+        const Real g3 = homsq * (SQR(zm) - SQR(zp)) / (zp - zm);
```
adamdempsey90: could just replace this with  `-om0*zc`

adamdempsey90 (commented):
Mostly minor things, looks good otherwise.

src/dust/dust.cpp
pdmullen: Yes.  Will do.

src/artemis.hpp
pdmullen: :+1:

src/rotating_frame/rotating_frame.hpp
pdmullen: :+1:

src/rotating_frame/rotating_frame.hpp
pdmullen: I guess this MR is missing a check on `nghost`.  If one selected `pcm` reconstruction for the residual equations with `nghost=1`, then linear advection would segfault.  If we switched to PPM for advection, we would segfault if `nghost=2`.  

Also, PPM reconstruction requires that I update the sweep logic.  Unfortunately PLM is kinda embedded in the current implementation.  A future implementation might set `rotating_frame/reconstruct=xxx`... but that would require some work.

src/rotating_frame/rotating_frame.hpp
pdmullen: :+1:

src/rotating_frame/rotating_frame_impl.hpp
pdmullen: Would that undo your efforts on symmetry?

src/rotating_frame/rotating_frame.hpp
pdmullen: The `jend` update is handled outside the loop.  But I'll recheck the logic...

src/rotating_frame/rotating_frame_impl.hpp
adamdempsey90: they should be equivalent. But we are not doing the symmetry preserving energy update anyway

src/rotating_frame/rotating_frame.hpp
adamdempsey90: I ask just because parthenon's par_for is `<=`

adamdempsey90: but i see now that you do that last index outside the loop

src/artemis.hpp
pdmullen: This produces some really ugly line breaks in the new upwinding infrastructure due to long lines.  Can't figure out where to otherwise shorten... hoping we can keep as is (but I will add null to this enum class and add left/right to comments above)

src/dust/dust.cpp
pdmullen: this is also a bug... needs to only operate on d == 1

src/rotating_frame/rotating_frame.hpp
pdmullen: Added check on nghost

src/rotating_frame/rotating_frame_impl.hpp
pdmullen: switched to simplified construction... only reason I kept around the previous formulation was because I thought it was the main contributor to your symmetry fix

src/gas/gas.cpp
```diff
@@ -421,13 +422,25 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
...
 
+  // NOTE(@pdmullen): Without FARGO, integration of the residual eqs must have a dt
+  // limited by vy = dvy + vy0 = dvy - q Omega x
+  bool do_shear = false;
+  Real qshear = 0.0, om0 = 0.0;
+  if (pm->packages.Get("artemis")->Param<bool>("do_rotating_frame")) {
+    auto &rframe_pkg = pm->packages.Get("rotating_frame");
+    qshear = rframe_pkg->Param<Real>("qshear");
+    om0 = rframe_pkg->Param<Real>("omega");
+    do_shear = (qshear != 0.0);
```
adamdempsey90:
```suggestion
    do_shear = (qshear*om0 != 0.0);
```

src/rotating_frame/rotating_frame.hpp
```diff
@@ -14,38 +14,294 @@
...
+//! \brief Returns signed shear velocity
+template <Coordinates GEOM>
+KOKKOS_INLINE_FUNCTION Real ShearVelocity(const Real qshear, const Real omega,
+                                          const Real x1v) {
+  if constexpr (GEOM == Coordinates::cartesian) {
+    return -qshear * omega * x1v;
+  } else {
+    PARTHENON_FAIL("Shearing box currently only supports Cartesian geometries");
+  }
+}
```
adamdempsey90:
```suggestion
template <Coordinates GEOM>
KOKKOS_INLINE_FUNCTION 
std::array<Real,3> BackgroundVelocity(const std::array<Real,3> &x, const Real qshear, const Real omega) {
  std::array<Real,3> vb{0.0};
  if constexpr (GEOM == Coordinates::cartesian) {
    vb[1] = -qshear * omega * x[0];
  } else if constexpr( GEOM == Coordinates::cylindrical) {
    vb[1] = omega * x[0];
  } else if constexpr( GEOM == Coordinates::axisymmetric) {
    vb[2] = omega * x[0];
  } else if constexpr( GEOM == Coordinates::spherical) {
    vb[2] = omega * x[0]*std::sin(x[1]);
  }
  return vb;
}
```
This is more what I was thinking. Just return a zero vector if this isn't active and subtract / add zeros in places where it's awkward to have a switch with `do_shear` (if such places exist).

src/rotating_frame/rotating_frame.hpp
adamdempsey90: Mainly pointing out for when we do the real fargo, we can keep plm for now

src/dust/dust.cpp
```diff
@@ -264,7 +277,10 @@ Real EstimateTimestepMesh(MeshData<Real> *md) {
         for (int n = 0; n < vmesh.GetSize(b, dust::prim::density()); ++n) {
           Real denom = 0.0;
           for (int d = 0; d < ndim; d++) {
-            denom += std::abs(vmesh(b, dust::prim::velocity(VI(n, d)), k, j, i)) / dx[d];
+            Real vd = vmesh(b, dust::prim::velocity(VI(n, d)), k, j, i);
+            vd += (do_shear && (d == 1)) ? ShearVelocity<GEOM>(qshear, om0, coords.x1v())
+                                         : 0.0;
+            denom += std::abs(vd) / dx[d];
```
adamdempsey90: I'm not sure you actually need to add the background anymore. The euler equations you're solving now are the ones with the residual velocity, so their signal speed is different. 

Your linear advection step does still need to limit the timestep such that  dt < dx/vb .

src/rotating_frame/rotating_frame_impl.hpp
adamdempsey90: That was mostly boundary conditions.

src/dust/dust.cpp
pdmullen: Agreed and addressed

src/rotating_frame/rotating_frame.hpp
pdmullen: This is overkill for now, since my edits in this MR only address the shearing box with `qshear != 0`.  Let's punt on this until we address more generally how to handle background flows.

src/gas/gas.cpp
pdmullen: :+1:

src/rotating_frame/rotating_frame.hpp
adamdempsey90: Let's compromise. I'm fine with removing the unused geometries but can we keep the return of the full 3D vector for the background velocity? The nbody package needs that full velocity in its sink method (even for shearing box) so it can add/subtract from the gas.

pdmullen: Addressed in [5410d15](https://github.com/lanl/artemis/pull/68/commits/5410d158af596b360df47bda43566f281782e55a)

adamdempsey90:
I think this is ready to go. We will deal with the method deficiencies in a future PR.

adamdempsey90 (approved):

PR #79, opened 2025-06-19
Draft: Subcycle linear advection

pdmullen:
## Checklist

- [ ] New features are documented
- [ ] Tests added for bug fixes and new features
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #80, opened 2025-06-22
Orbital advection for shearing box

adamdempsey90:
## Background

Replaces the upwind advection with an exact remap. Also incorporates #79 for larger time steps.

Appears to work, but more testing is needed.

There is also a question about whether we are doing more than we need to with the `P2C` and `C2P` tasks after the remap. We should get that bit streamlined before merging

## Checklist


src/artemis_driver.cpp
```diff
@@ -70,7 +70,13 @@ ArtemisDriver<GEOM>::ArtemisDriver(ParameterInput *pin, ApplicationInput *app_in
   do_imc = artemis_pkg->template Param<bool>("do_imc");
   do_moment = artemis_pkg->template Param<bool>("do_moment");
 
-  // Moments integrator and initialization
+  // Linear advection integrator
+  // NOTE(@pdmullen): Hardcoded to rk2 until replaced with exact remap
```
adamdempsey90:
```suggestion
```

src/gas/gas.cpp
```diff
@@ -61,25 +61,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
...
-                      "PPM requires at least 3 ghost cells.");
-    if (coords != Coordinates::cartesian) {
-      PARTHENON_WARN("Artemis' PPM implementation does not contain geometric corrections "
-                     "for curvilinear coordinates.");
-    }
-    recon_method = ReconstructionMethod::ppm;
-  } else {
-    PARTHENON_FAIL("Reconstruction method not recognized.");
-  }
+  recon_method = ArtemisUtils::ChooseReconMethod(recon);
```
pdmullen, line 64: Use this new utility in dust package.

pdmullen: Fixed

pdmullen (approved):

PR #91, opened 2025-08-21
Coagulation Refactor

pdmullen:
## Checklist


src/dust/coagulation/coagulation.cpp
```diff
@@ -94,223 +90,289 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin, Params &dustPar
...
+  params.Add("coag_scr_level",
+             pin->GetOrAddInteger("dust/coagulation", "coag_scr_level", 0));
+
+  // Fields for stashing solver diagnostics
   const bool info_out = pin->GetOrAddBoolean("dust/coagulation", "coag_info_out", false);
   params.Add("coag_info_out", info_out);
+  if (info_out) {
+    Metadata m = Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy});
+    coag->AddField<dust::coag::ncalls>(m);
+  }
```
adamdempsey90, line 144: Is this really needed?

src/dust/coagulation/coagulation.cpp
```diff
@@ -94,223 +90,289 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin, Params &dustPar
...
+        const auto &xv = coords.GetCellCenter();
+        const auto &xcyl = coords.ConvertToCyl(xv);
+        const Real irad = coag.const_omega ? 1.0 : 1.0 / xcyl[0]; // cylindrical
+        const Real omega1 = irad * std::sqrt(irad) / time0;       // code-units
+
+        // Extract gas state vector
+        const Real &gdens = vmesh(b, gas::prim::density(0), k, j, i);
+        const Real &gsie = vmesh(b, gas::prim::sie(0), k, j, i);
+        const Real &gbulk = eos_d.BulkModulusFromDensityInternalEnergy(gdens, gsie);
+        const Real cs1 = std::sqrt(gbulk / gdens) * vel0;
```
adamdempsey90, line 296: I think the right thing to do with this is the thermal speed 
$v_{\rm th}^2 = (8/\pi) k_B T$
instead of sound speed

src/dust/coagulation/coagulation.cpp
```diff
@@ -94,223 +90,289 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin, Params &dustPar
...
+                const int vidx = VI(n, d);
+                vmesh(b, dust::cons::momentum(vidx), k, j, i) =
+                    gt0 * rhod(n) * vel(vidx) * hx[d] / (rho0 * vel0);
+              }
+            });
+
+        // Diagnostics
+        if (info_out_flag) {
+          Kokkos::single(Kokkos::PerTeam(mbr),
+                         [&]() { vmesh(b, dust::coag::ncalls(), k, j, i) = ncall; });
```
adamdempsey90, line 343: so the conclusion is that the threads used in the inner loops are active in the outer?

src/dust/coagulation/coagulation.hpp
```diff
@@ -172,25 +502,24 @@ Real CoagulationRate(const int i, const int j, const Real kernel4[],
   const Real mu = 2.3;            //! mean molecular mass in proton masses
   const Real m_p = 1.6726231e-24; //! proton mass in g
```
adamdempsey90, line 506: `mu` is in our gas description and stored in the EOS I think. And we should pull proton mass from our constants

src/dust/coagulation/coagulation.hpp
```diff
@@ -172,25 +502,24 @@ Real CoagulationRate(const int i, const int j, const Real kernel4[],
...
+  // Calculate some basic properties
   const Real hg = cs / omega;
   Real re = alpha * sig_h2 * gasdens / (2.0 * mu * m_p);
-  if (coag.coord == 0) {
-    re *= std::sqrt(2.0 * M_PI) * hg;
-  }
+  if (!(coag.coord)) re *= std::sqrt(2.0 * M_PI) * hg;
 
   const Real tn = 1.0 / omega;
   const Real ts = tn / sqrt(re);
```
adamdempsey90, line 514:
```suggestion
  const Real ts = tn / std::sqrt(re);
```

src/dust/coagulation/coagulation.hpp
```diff
@@ -441,21 +763,21 @@ void Coagulation_nQ(parthenon::team_mbr_t const &mbr, ScratchPad1D<Real> &nQs,
...
       const Real Rf1 = distri(i) * distri(j) * fett_l;
-      const Real dummy = coag.coagR3D(iepsFrag, i, j) * Rf1 * Q(i);
+      const Real dummy = coag.coagR3D(iepsfrag, i, j) * Rf1 * Q(i);
       sum0 += dummy;
     }
 
-    if (i - pGrid - 1 >= 0) {
+    if (i - pgrid - 1 >= 0) {
       Kokkos::atomic_add(&nQs(i - 1), sum0);
     }
```
adamdempsey90, line 779: We could consider `ScatterView`

src/dust/coagulation/coagulation.cpp
pdmullen: This change transitions from `ncalls` being a ParArray3D to a field.  There might be a way to make the main coagulation kernel's outer loop a reduction over ncalls... but I already have concern about thread safety of that particular kernel.

src/dust/coagulation/coagulation.cpp
pdmullen: I worried here about multiple threads with shared `kji` setting `ncalls`.

src/dust/coagulation/coagulation.cpp
pdmullen: :+1:

src/dust/coagulation/coagulation.hpp
pdmullen: Definitely.

src/dust/coagulation/coagulation.hpp
pdmullen: nice catch!

src/dust/coagulation/coagulation.hpp
pdmullen: You are learning alot of Kokkos.

src/dust/coagulation/coagulation.cpp
pdmullen: I want to make sure we are using the right speed in the right place.  Can you take a look here:

https://github.com/lanl/artemis/blob/pdmullen/coag-v2/src/dust/coagulation/coagulation.hpp#L516

where it appears we actually invoke the speed you give above.

src/dust/coagulation/coagulation.cpp
adamdempsey90: I meant do we want to remove ncalls entirely?

src/dust/coagulation/coagulation.cpp
adamdempsey90: I think this all gets back to the question of do we want to write this to be PPD specific or more generic where something like $H = c_s / \Omega$ is not a thing? There is a mixture of sound speed usage that is real (e.g., a Mach number) or PPD specific (e.g., scale height) or ideal gas specific (e.g., approximating the thermal speed).

src/dust/coagulation/coagulation.hpp
adamdempsey90: It's being forced upon me

PR #92, opened 2025-08-29
Dust Coagulation v2

pdmullen:
## Checklist


src/artemis.cpp
```diff
@@ -114,6 +116,8 @@ Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
                     "Radiation requires the gas package, but there is not gas!");
   PARTHENON_REQUIRE(!(do_imc && do_moment),
                     "Cannot simultaneously evolve IMC and moments radiation");
+  PARTHENON_REQUIRE(!(do_coagulation) || (do_coagulation && do_dust),
+                    "Coagulation requires the dust package, but there is not dust!");
```
pdmullen, line 125: Per team discussion, require `do_gas` for coagulation.

src/dust/coagulation/coagulation.cpp
```diff
@@ -0,0 +1,382 @@
...
+  params.Add("nstep_coag", pin->GetOrAddInteger("dust/coagulation", "nstep_coag", 50));
+  params.Add("dt_coag", 0.0, Params::Mutability::Restart);
+  params.Add("coag_alpha", pin->GetOrAddReal("dust/coagulation", "coag_alpha", 1.e-3));
+  params.Add("coag_scr_level",
+             pin->GetOrAddInteger("dust/coagulation", "coag_scr_level", 0));
+
+  // Fields for stashing solver diagnostics
+  const bool info_out = pin->GetOrAddBoolean("dust/coagulation", "coag_info_out", false);
+  params.Add("coag_info_out", info_out);
+  if (info_out) {
```
pdmullen: Should we keep this? --- From #91 @adamdempsey90

src/dust/coagulation/coagulation.cpp
```diff
@@ -0,0 +1,382 @@
...
+        const auto &xv = coords.GetCellCenter();
+        const auto &xcyl = coords.ConvertToCyl(xv);
+        const Real irad = coag.const_omega ? 1.0 : 1.0 / xcyl[0]; // cylindrical
+        const Real omega1 = irad * std::sqrt(irad) / time0;       // code-units
+
+        // Extract gas state vector
+        const Real &gdens = vmesh(b, gas::prim::density(0), k, j, i);
+        const Real &gsie = vmesh(b, gas::prim::sie(0), k, j, i);
+        const Real &gbulk = eos_d.BulkModulusFromDensityInternalEnergy(gdens, gsie);
+        const Real cs1 = std::sqrt(gbulk / gdens) * vel0;
```
pdmullen, line 311: Sound speed vs thermal speed? --- From #91 @adamdempsey90

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1045 @@
...
+  const Real &cs = kernel4[2];
+  const Real &omega = kernel4[3];
+  const Real &tau_i = stoppingTime(i);
+  const Real &tau_j = stoppingTime(j);
+  const Real *vel_i = &vel(3 * i);
+  const Real *vel_j = &vel(3 * j);
+
+  const Real sig_h2 = 2.0e-15;    //! cross section of H2
+  const Real mu = 2.3;            //! mean molecular mass in proton masses
+  const Real m_p = 1.6726231e-24; //! proton mass in g
```
pdmullen: This is not general, and may even clash with definitions in gas package.  Share `mu` among packages.  Extract `m_p` from units system, and perhaps make cross section a user settable param.  --- Derived from #91 @adamdempsey90

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1045 @@
...
+    Real sum0 = 0.0;
+    for (int j = 0; j <= i - pgrid - 1; j++) {
+      const Real fett_l = CoagulationRate(i, j, kernel4, vel, stoppingTime, coag, 1);
+      const Real Rf1 = distri(i) * distri(j) * fett_l;
+      const Real dummy = coag.coagR3D(iepsfrag, i, j) * Rf1;
+      sum0 += dummy;
+    }
+
+    if (i - pgrid - 1 >= 0) {
+      Kokkos::atomic_add(&source(i - 1), sum0);
```
pdmullen, line 449: `ScatterView` --- From #91 @adamdempsey90

src/dust/coagulation/coagulation.cpp
```diff
@@ -0,0 +1,382 @@
...
+  using namespace ::parthenon::Update;
+  const int num_partitions = pm->DefaultNumPartitions();
+  TaskRegion &tr = tc.AddRegion(num_partitions);
+  for (int i = 0; i < num_partitions; i++) {
+    auto &tl = tr[i];
+    auto &base = pm->mesh_data.GetOrAdd("base", i);
+    auto &md_coag = pm->mesh_data.GetOrAdd("md_coag", i);
+
+    // Execute coagulation step
+    auto coag_step = tl.AddTask(none, CoagulationStep<GEOM>, base.get(), ltime, ldt);
```
adamdempsey90, line 199: A lot happens inside of this task, are there opportunities for splitting this up into multiple tasks?

pdmullen: This task launches only one hierarchical loop (as currently written) when not outputting coagulation diagnostics. 

Maybe we could separate the coagulation diagnostics into their own tasks, but since they are already hidden behind an `info_out` flag, I don't think the current single task is too bad. 

In [0a5d72b](https://github.com/lanl/artemis/pull/92/commits/0a5d72b3545ef3ce485e428101c1e59eab734a70), I did remove the `ncalls` reduction which permitted me to replace `PreCoagulationDiagnostics` and `PostCoagulationDiagnostics` with a single `CoagulationDiagnostics`.

src/dust/coagulation/coagulation.cpp
pdmullen: @shengtai can you advise here?

shengtai: The coagulation algorithm fromTil  Birnstiel uses the sound speed to calculate the temperature in their code. The sound speed should be used here. I am not sure what is the difference between thermal speed and sound speed

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,994 @@
...
+  Real time_goal = time_dummy + dt_sync1;
+  Real dt = dt_sync1;
+  Real hnext = dt;
+  dt_sync = 1e-15; // works H5
+
+  Real kernel[4];
+  kernel[0] = gdens;
+  kernel[1] = alpha;
+  kernel[2] = cs;
+  kernel[3] = omega;
```
adamdempsey90: This, and similar things, should become a struct of these parameters.

src/dust/coagulation/coagulation.cpp
adamdempsey90: If the sound speed is used to compute temperature, we should just pass the temperature. If the sound speed is used for calculating a scale height, we should reformulate that to not use the scale height. If the sound speed is used to calculate a viscosity, we should pass the viscosity structs instead. 

We should only be using this sound speed of something actually needs it. If anything uses PPD quantities (surface density, scale height, alpha viscosity, etc.), we need to replace them with their more generic counterparts (e.g., $\Sigma/H$ is replaced with $\rho$).

adamdempsey90: If the sound speed is used to compute temperature, we should just pass the temperature. If the sound speed is used for calculating a scale height, we should reformulate that to not use the scale height. If the sound speed is used to calculate a viscosity, we should pass the viscosity structs instead. 

We should only be using this sound speed of something actually needs it. If anything uses PPD quantities (surface density, scale height, alpha viscosity, etc.), we need to replace them with their more generic counterparts (e.g., $\Sigma/H$ is replaced with $\rho$).

shengtai: I checked their coagulation kernel and found c_s (sound speed) is used to calculate the scale height, temperature, relative velocities, etc. We have two options: pass all informations to coagulation kernel: sound speed, scale height, and temperature using our calculated value, or pass only sound speed and let coagulation kernel to calculate other info needed.  I prefer to pass only the sound speed.  

To my knowledge, Til's algorithm is designed for proto-planetary disk applications. That is why it needs alpha, omega, and scale height etc. Its temperature calculation may also differ our temperature calculation in artemis. If we need to rewrite coagulation for general purpose (not only for disk applications), we can change their algorithm for other applications.

src/dust/coagulation/coagulation.hpp
pdmullen: I'll do this one...

pdmullen: [4ee353a](https://github.com/lanl/artemis/pull/92/commits/4ee353a9c2edc40f5d9b194a0bbdc29f650c0d8d)

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+        }
+      });
+
+  // Set fragmentation variables
+  const Real ten_a = std::pow(10.0, a);
+  const Real ten_ma = 1.0 / ten_a;
+  const int ce = int(-1.0 / a * std::log10(1.0 - ten_ma)) + 1;
+
+  // Used in integration
+  pgrid = floor(1.0 / a);
```
adamdempsey90:
```suggestion
  const int ce = static_cast<int>( std::floor(-1.0 / a * std::log10(1.0 - ten_ma)))  + 1;

  // Used in integration
  pgrid = static_cast<int>(std::floor(1.0 / a));
```

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+
+  // Initialization Part II
+  const int iphifrag = coag2drv::phifrag;
+  const int iepsfrag = coag2drv::epsfrag;
+  const int iafrag = coag2drv::afrag;
+  const Real frag_slope = 2.0 - 11.0 / 6.0;
```
adamdempsey90:
```suggestion
  const Real frag_slope = 1.0/6.0; // = 2.0 - 11.0 / 6.0;
```

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+      0, nm - 1, KOKKOS_LAMBDA(const int i) {
+        Real sum_pF = 0.0;
+        for (int j = 0; j <= i; j++) {
+          coag3d(iphifrag, j, i) = std::pow(mass_grid(j), frag_slope);
+          sum_pF += coag3d(iphifrag, j, i);
+        }
+        // normalization
+        for (int j = 0; j <= i; j++) {
+          coag3d(iphifrag, j, i) /= sum_pF; // switch (i,j) from fortran
+        }
```
adamdempsey90: I'm seeing a mix of `(j,i)` and `(i,j)` usage throughout this file. Should there be one common indexing pattern, or do you really want to reverse the order for some?

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+      const Real fett_t = CoagulationRate(i, j, kernel4, vel, stoppingTime, coag, 0);
+      const Real Rc1 = distri(i) * distri(j) * fett_t;
+      for (int nz = 0; nz < 4; nz++) {
+        const int k = coag.cpod_notzero(i, j, nz);
+        if (k < 0) continue;
+        const Real src_coag0 = coag.cpod_short(i, j, nz) * Rc1;
+        Kokkos::atomic_add(&source(k), src_coag0);
+      }
+    }
+  });
```
adamdempsey90, line 370: The atomic can be replaced with an inner reduce?

src/dust/coagulation/coagulation.hpp
adamdempsey90: The `i` thread is updating the `i-1` thread with the sum it created. Can this `i` loop be refactored so that thread `i` updates `source(i)` instead? I think the answer is yes and you just replace `i` with `i+1` in the sum calculation above?

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+    Real sum1 = -sum0;
+    // Full fragmentation (only negative terms)
+    int i1 = std::max(0, i - pgrid);
+    for (int j = i1; j <= i; j++) {
+      const Real fett_l = CoagulationRate(i, j, kernel4, vel, stoppingTime, coag, 1);
+      const Real Rf1 = distri(i) * distri(j) * fett_l;
+      sum1 -= Rf1;
+    }
+    Kokkos::atomic_add(&source(i), sum1);
+  });
```
adamdempsey90:
```suggestion
    Real sum1 = -sum0;
    // Full fragmentation (only negative terms)
    int i1 = std::max(0, i - pgrid);
    for (int j = i1; j <= i; j++) {
      const Real fett_l = CoagulationRate(i, j, kernel4, vel, stoppingTime, coag, 1);
      const Real Rf1 = distri(i) * distri(j) * fett_l;
      sum1 -= Rf1;
    }
    source(i) += sum1;
  });
```
if the loop above this is modified as suggested

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+                    const ScratchPad1D<Real> &vel, const ScratchPad1D<Real> &stoppingTime,
+                    const CoagParams &coag) {
+  // Adding coagulation source terms
+  const int iafrag = coag2drv::afrag;
+  const int iphifrag = coag2drv::phifrag;
+  const int iepsfrag = coag2drv::epsfrag;
+  const int idalp = coag2drv::dalp;
+  const int idpod = coag2drv::dpod;
+  parthenon::par_for_inner(DEFAULT_INNER_LOOP_PATTERN, mbr, 0, mimax, [&](const int i) {
+    for (int j = 0; j <= i; j++) {
```
adamdempsey90, line 493: could this be a 2D `par_for_inner` with a screen on `j <= i`?

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+      const Real tmp2 = dphjiQ - Qp1 * dpodij;
+      for (int nz = 0; nz < 4; nz++) {
+        const int k = coag.cpod_notzero(i, j, nz);
+        if (k < 0) continue;
+        const Real kdeltajk = (j == k) ? 1.0 : 0.0;
+        const Real kdeltaik = (i == k) ? 1.0 : 0.0;
+        const Real nqs_k =
+            (Qp1 * coag.cpod_short(i, j, nz) + tmp1 * kdeltajk + tmp2 * kdeltaik) * Rc1;
+        Kokkos::atomic_add(&nQs(k), nqs_k);
+      }
```
adamdempsey90, line 518:
```suggestion
      Real nqs_k = 0.0;
      for (int nz = 0; nz < 4; nz++) {
        const int k = coag.cpod_notzero(i, j, nz);
        if (k < 0) continue;
        const Real kdeltajk = (j == k) ? 1.0 : 0.0;
        const Real kdeltaik = (i == k) ? 1.0 : 0.0;
        nqs_k +=
            (Qp1 * coag.cpod_short(i, j, nz) + tmp1 * kdeltajk + tmp2 * kdeltaik) * Rc1;
      }
      Kokkos::atomic_add(&nQs(k), nqs_k);
```
This one should really be a `ScatterView`, but I don't know if you can do that for scratchpads?

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+
+    Real sum1 = -sum0;
+    // Full fragmentation (only negative terms)
+    int i1 = std::max(0, i - pgrid);
+    for (int j = i1; j <= i; j++) {
+      const Real fett_l = CoagulationRate(i, j, kernel4, vel, stoppingTime, coag, 1);
+      const Real Rf1 = distri(i) * distri(j) * fett_l * Q(i);
+      sum1 -= Rf1;
+    }
+    Kokkos::atomic_add(&nQs(i), sum1);
```
adamdempsey90: should be able to remove all atomics by incrementing the `i` used in the first sum calculation

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+        [&](const int i, int &lmax) {
+          if (dustdens(i) > coag.dfloor / coag.mass_grid(i)) {
+            lmax = std::max(lmax, i);
+          }
+        },
+        Kokkos::Max<int>(mimax));
+
+    CoagulationSource(mbr, source, dustdens, mimax, kernel, vel, stime, coag);
+
+    if (coag.use_adaptive == 0 || coag.integrator == 1) {
```
adamdempsey90: Can some of these options be pulled out of this while loop done for every cell and made into a template parameter for this function?

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,1000 @@
...
+      }
+
+    } else {
+      // third-order method
+      Real h0 = hnext;
+      Real h = h0;
+      Real errmax;
+      int mimax2 = 0;
+      // Heun's Method
+      while (1) {
```
adamdempsey90, line 909: Is there a sensible maximum number of iterations that we can put here to avoid an infinite loop?

adamdempsey90 (commented):
Went through `coagulation.hpp`

src/dust/coagulation/coagulation.cpp
adamdempsey90: In the coagulation paper, the temperature and c_s are used for the brownian relative velocity calculation. So I've added both here.

adamdempsey90:
@pdmullen Can I trouble you for one last look through to see if there's anything major that we've missed.

adamdempsey90 (approved):

tst/scripts/coagulation/__init__.py
RyanWollaeger, line 1: Is this an empty file?

src/pgen/coag.hpp
```diff
@@ -0,0 +1,168 @@
+//========================================================================================
+// (C) (or copyright) 2023-2024. Triad National Security, LLC. All rights reserved.
```
RyanWollaeger: Update copyright range on these new files?

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,968 @@
...
+                        mass_grid, coagR3D, chi, surface, rate);
+  CrateringSourceNQ(mbr, nm1, mimax, Q, nQs, dustdens, vel, stime, kernel, mass_grid,
+                    coagR3D, surface, rate);
+  FinalizeSourceNQ(mbr, nm1, mimax, pgrid, Q, nQs, dustdens, vel, stime, kernel,
+                   mass_grid, coagR3D, surface, rate);
+}
+
+//----------------------------------------------------------------------------------------
+//! \fn  void Dust::Coagulation::
+//  \brief
```
RyanWollaeger, line 820: Worth adding a brief description here and in the source functions above?

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,968 @@
...
+KOKKOS_INLINE_FUNCTION
+void SourceNQ(const parthenon::team_mbr_t &mbr, const int &n, const int &nm1,
+              const int &mimax, const int &pgrid, const ScratchPad1D<Real> &Q,
+              const ScratchPad1D<Real> &nQs, const ScratchPad1D<Real> &dustdens,
+              const ScratchPad1D<Real> &vel, const ScratchPad1D<Real> &stime,
+              const StateParams &kernel, const ParArray1D<Real> &mass_grid,
+              const ParArray3D<Real> &coagR3D, const ParArray3D<int> &Kijk_sym_ind,
+              const ParArray3D<Real> &Kijk_sym, const Real &chi, const bool &surface,
+              const RateParams &rate) {
+  InitializeSourceNQ(mbr, nm1, mimax, Q, nQs, dustdens, vel, stime, kernel, mass_grid,
```
RyanWollaeger, line 808: Just curious - why does this `SourceNQ` not have a `ZeroSourceNQ`? Otherwise it looks analogous to the `Source` routine above (sorry if missing something - don't know this code/physics really)

src/dust/coagulation/coagulation.hpp
```diff
@@ -0,0 +1,968 @@
...
+
+    int mimax2 = Null<int>();
+    if (!(do_adaptive) || (coag_int == 1)) {
+      dt_sync1 = TimeStepControl(mbr, nm1, dustdens, source, mass_grid, dfloor, cfl);
+      dt = std::min(dt_sync1, time_goal - time_dummy);
+      dt_sync = dt_sync1;
+      if (coag_int == 3) {
+        mimax2 = FindMIMaxNQS3(mbr, nm1, dt, dustdens, source, Q, mass_grid, dfloor);
+        // now Q stores dustdens + dt*source(), nQs will be used for temperary source(*)
+        ZeroSource(mbr, nm1, nQs);
```
RyanWollaeger, line 901: Is this `ZeroSource` (and the one below) redundant with the one inside `Source`? (Again - don't know what is going on - feel free to ignore)

RyanWollaeger (commented):
Sorry if these questions/comments are not constructive.

src/dust/coagulation/coagulation.hpp
adamdempsey90: `ZeroSourceNQ` is called before `SourceNQ` rather than being called within it. So I think it's fine.

src/pgen/coag.hpp
adamdempsey90: Done

src/dust/coagulation/coagulation.hpp
adamdempsey90: I've added some descriptions, but left others as the function name is descriptive enough.

tst/scripts/coagulation/__init__.py
adamdempsey90: Yes, all of these `__init__.py` files are empty

RyanWollaeger (approved):

PR #97, opened 2025-10-01
Fix bug in disk

adamdempsey90:
## Background
This has turned into changing the way we set the azimuthal velocity in the disk problem. Before, we evaluated the pressure on all of the faces of a cell, then computed the pressure gradient in the problem coordinate system, then converted that to dP/dR in cylindrical coordinates. This MR changes this to simply evaluating the pressure profile at +/- epsilon of the cells centroid R value. In hindsight, this is much simpler and probably more robust. 

In testing, I noticed that the code is sensitive to the density floor value -- it can't be too low. I've gone through and updated all of the example disk input files to use a floor of 1e-10. Some inputs also needed updating because they were using out of date nodes/options.

Other features I've added:
- The floor values for pressure and the new temperature floor are now set consistently from the density and sie floors of gas. 
- The extrapolation BC allows outflow but not inflow.
- User can control power of exponential cut off. 

## Checklist


pdmullen (approved):

adamdempsey90:
Sorry @pdmullen, looks like this was previously set to auto-merge when CI passes, but the CI was failing. When I updated the branch with the new stuff, the tests starting passing and it got merged. Happy to make any other modifications in another MR

PR #105, opened 2025-11-21
Fix restarts with IMC (simply avoid auto-inputblock enrollment).

RyanWollaeger:
## Background

* Longer running problems require restarts.
* IMC is the dominant cost when active, further enhancing the need for restarts.

## Description of Changes

+ Check !do_imc when invoking GetOrAdd radiation/moments parameter.
+ Use xor to ensure at least one of moments or IMC with radiatioin.
+ Add plot.py options: to show plot, use data bounds in colorbar.

## Checklist


pdmullen (approved):

src/artemis.cpp
```diff
@@ -114,8 +114,8 @@ Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
                     "Conduction requires the gas package, but there is not gas!");
   PARTHENON_REQUIRE(!(do_radiation) || (do_radiation && do_gas),
                     "Radiation requires the gas package, but there is not gas!");
-  PARTHENON_REQUIRE(!(do_imc && do_moment),
-                    "Cannot simultaneously evolve IMC and moments radiation");
+  PARTHENON_REQUIRE(do_radiation ? (do_moment ^ do_imc) : !(do_moment || do_imc),
+                    "Radiation requires exactly one of moment or IMC method");
```
adamdempsey90: XOR is too complicated for my simple brain

adamdempsey90: Can you write this in a simpler way? I'm fine with some if-else branches if that makes this easier to understand

RyanWollaeger: To XOR XOR not to XOR, that was the question.

RyanWollaeger: Split this ternary into separate conditions and removed the XOR. The user messages are more specific now since they correspond to the distinct input possibilities, which is nice.

adamdempsey90 (approved):
Thanks for catching this

PR #118, opened 2026-03-20
Turn on/off radiation

adamdempsey90:
## Background

This adds the machinery to turn on/off and initialize/shutdown a package during a simulation. To start, I have enrolled moments and raytrace into this. 

When moments comes on, it will set all variables to LTE values.

We should do the same for IMC and pretty much all of the packages.

I have tested this locally, and it works correctly, even through restarts.

## Checklist

### DIFF
--- a/src/radiation/imc/imc_driver.cpp
+++ b/src/radiation/imc/imc_driver.cpp
@@ -37,16 +37,16 @@ TaskListStatus JaybenneIMC(Mesh *pmesh, const SimTime &tm, const Real dt) {
     return TaskListStatus::complete;
   } else if (active == ArtemisUtils::PackageControl::shutdown) {
     if (Globals::my_rank == 0) {
-      printf("Turning off iMC radiation at t=%.8e...\n", tm.time);
+      printf("Turning off IMC radiation at t=%.8e...\n", tm.time);
     }
     return TaskListStatus::complete;
   } else if (active == ArtemisUtils::PackageControl::initial) {
     if (Globals::my_rank == 0) {
       printf("Turning on IMC radiation at t=%.8e...\n", tm.time);
     }
-    // What to call here to set the field?
-    // Need a pmesh version of this
-    // jaybenne::InitializeRadiation(md.get(), true)
+    for (auto &mbd : pmesh->mesh_data.Get()->GetAllBlockData()) {
+      jaybenne::InitializeRadiation(mbd.get(), true);
+    }
   }
   auto status = Radiation::UpdateRadiationFields(pmesh).Execute();
   if (status != TaskListStatus::complete) return status;
--- a/src/radiation/imc/imc_driver.cpp
+++ b/src/radiation/imc/imc_driver.cpp
@@ -16,6 +16,7 @@
 #include "derived/fill_derived.hpp"
 #include "radiation/imc/imc.hpp"
 #include "radiation/radiation.hpp"
+#include "utils/artemis_utils.hpp"
 
 // Jaybenne includes
 #include "jaybenne.hpp"
@@ -30,6 +31,23 @@ namespace IMC {
 template <Coordinates GEOM>
 TaskListStatus JaybenneIMC(Mesh *pmesh, const SimTime &tm, const Real dt) {
   PARTHENON_INSTRUMENT
+  auto &pkg = pmesh->packages.Get("radiation");
+  const auto active = ArtemisUtils::CheckPackageStatus(pkg, tm.time);
+  if (active == ArtemisUtils::PackageControl::inactive) {
+    return TaskListStatus::complete;
+  } else if (active == ArtemisUtils::PackageControl::shutdown) {
+    if (Globals::my_rank == 0) {
+      printf("Turning off iMC radiation at t=%.8e...\n", tm.time);
+    }
+    return TaskListStatus::complete;
+  } else if (active == ArtemisUtils::PackageControl::initial) {
+    if (Globals::my_rank == 0) {
+      printf("Turning on IMC radiation at t=%.8e...\n", tm.time);
+    }
+    // What to call here to set the field?
+    // Need a pmesh version of this
+    // jaybenne::InitializeRadiation(md.get(), true)
+  }
   auto status = Radiation::UpdateRadiationFields(pmesh).Execute();
   if (status != TaskListStatus::complete) return status;
   status = jaybenne::RadiationStep(pmesh, tm, dt).Execute();
--- a/src/radiation/moments/moments.cpp
+++ b/src/radiation/moments/moments.cpp
@@ -87,10 +87,12 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin,
              pin->GetOrAddBoolean("radiation/moment", "fatal_if_unconverged", true));
 
   // how to handle the matter coupling:
-  // full_coupling = false only does a loop over energy couopling
+  // full_coupling = false only does a loop over energy coupling
   // full_coupling = true also does an outer loop over momentum coupling
   params.Add("full_coupling",
              pin->GetOrAddBoolean("radiation/moment", "full_coupling", true));
+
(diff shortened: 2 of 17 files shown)


src/radiation/moments/moments.cpp
```diff
@@ -378,6 +380,52 @@ TaskStatus MatterCoupling(MeshData<Real> *u0, const Real dt) {
...
+    parthenon::par_for(
+        DEFAULT_LOOP_PATTERN, "Moments::InitMesh", DevExecSpace(), 0, md->NumBlocks() - 1,
+        kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+        KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+          // Extract coordinates
+          const Real &rho = vmesh(b, gas::prim::density(0), k, j, i);
+          const Real &sie = vmesh(b, gas::prim::sie(0), k, j, i);
+          const Real T = eos_d.TemperatureFromDensityInternalEnergy(rho, sie);
+          const Real Erad = arad * SQR(SQR(T));
+          vmesh(b, rad::cons::energy(0), k, j, i) = Erad;
```
RyanWollaeger: Would this be redundant with setup in some `pgen` files now, or should a user still initialize this data?

RyanWollaeger (approved):
LGTM

src/radiation/moments/moments.cpp
adamdempsey90: They should still initialize radiation data in the pgen and if tstart > 0, it just won't get used.

src/utils/artemis_utils.hpp
```diff
@@ -251,6 +251,44 @@ Real CutCell2D(const std::array<Real, 4> &x, const std::array<Real, 4> &y,
   return vol_inside / vol;
 }
 
+enum class PackageControl { inactive = -1, shutdown = 0, active = 1, initial = 2 };
+
+template <class PKG>
+inline PackageControl CheckPackageStatus(PKG &pkg, const Real time) {
```
pdmullen, line 257: Not a problem for today... but something like this might be a bit more nuanced for non operator-split physics.  For example, I don't think we would want to turn off physics in between the the 1st and 2nd stage of RK2.  Not an issue at the moment, just thinking aloud.

src/radiation/moments/moments.cpp
pdmullen: I wonder if we should mask such initialization on material with optical depth >~ 1>?

pdmullen (approved):

src/utils/artemis_utils.hpp
adamdempsey90: This is only called before entering `Step`, so there won't be any switching mid RK-stage

src/radiation/moments/moments.cpp
adamdempsey90: I worry that a batch of zones could skirt whatever criterion you impose and then you'd get bad numerical behavior of neighboring zones at the floor and not at the floor.

adamdempsey90: Will add `(do_this) ? std::min(tau, 1) * Erad : Erad;`

111-1000 (approved):

adamdempsey90:
I will add this machinery also to the cooling package

src/radiation/imc/imc_driver.cpp
```diff
@@ -30,6 +31,23 @@ namespace IMC {
...
+    }
+    return TaskListStatus::complete;
+  } else if (active == ArtemisUtils::PackageControl::initial) {
+    if (Globals::my_rank == 0) {
+      printf("Turning on IMC radiation at t=%.8e...\n", tm.time);
+    }
+    // What to call here to set the field?
+    // Need a pmesh version of this
+    // jaybenne::InitializeRadiation(md.get(), true)
+  }
```
adamdempsey90: If jaybenne had a mesh version of this function in addition to the meshblock version, we could turn imc on/off easily. @RyanWollaeger  @pdmullen let's add that to jaybenne

pdmullen: :+1:

src/self_gravity/poisson_driver.cpp
```diff
@@ -41,11 +42,26 @@ namespace SelfGravity {
 //----------------------------------------------------------------------------------------
 //! \fn TaskListStatus SelfGravity::PoissonDriver
 //! \brief
-void SolvePoisson(TaskCollection &tc, Mesh *pmesh) {
+void SolvePoisson(TaskCollection &tc, Mesh *pmesh, const Real time, const int stage) {
```
adamdempsey90, line 45: I noticed this doesn't return `TaskListStatus`, but should it @pdmullen ?

src/utils/artemis_utils.hpp
adamdempsey90: I have expanded this use to physics that is done in the rk substeps, in those cases i also only turn on/off for the first stage

src/self_gravity/poisson_driver.cpp
pdmullen: I think not.  All we are doing is passing the `TaskCollection tc` to `SolvePoisson` and having that function add a region with its own tasks. 

When we call `Execute` on the final `TaskCollection`, that returns `TaskListStatus` (see `StepTasks.Execute()`).  

The comment is wrong in the `//! \fn` though.

src/utils/artemis_utils.hpp
adamdempsey90: I have expanded this use to physics that is done in the rk substeps, in those cases i also only turn on/off for the first stage

src/radiation/moments/moments.cpp
adamdempsey90: Done

src/radiation/moments/moments_driver.cpp
```diff
@@ -28,6 +28,21 @@ TaskListStatus MomentsDriver(Mesh *pmesh, const SimTime &tm,
...
+  } else if (active == ArtemisUtils::PackageControl::shutdown) {
+    if (Globals::my_rank == 0) {
+      printf("Turning off radiation moments at t=%.8e...\n", tm.time);
+    }
+    return TaskListStatus::complete;
+  } else if (active == ArtemisUtils::PackageControl::initial) {
+    if (Globals::my_rank == 0) {
+      printf("Turning on radiation moments at t=%.8e...\n", tm.time);
+    }
+    Moments::InitMesh<GEOM>(pmesh);
```
pdmullen, line 44: Does this need to go over `IndexDomain::entire`? Or do we need to sneak in a call to boundary comms before proceeding?

src/radiation/moments/moments.cpp
```diff
@@ -378,6 +381,80 @@ TaskStatus MatterCoupling(MeshData<Real> *u0, const Real dt) {
...
+    // Packing and Indexing
+    static auto desc =
+        MakePackDescriptor<gas::prim::density, gas::prim::sie, rad::cons::energy,
+                           rad::prim::energy, rad::cons::flux, rad::prim::flux>(
+            (pmesh->resolved_packages).get());
+    auto vmesh = desc.GetPack(md);
+
+    IndexRange ib = md->GetBoundsI(IndexDomain::interior);
+    IndexRange jb = md->GetBoundsJ(IndexDomain::interior);
+    IndexRange kb = md->GetBoundsK(IndexDomain::interior);
```
adamdempsey90:
```suggestion
    IndexRange ib = md->GetBoundsI(IndexDomain::entire);
    IndexRange jb = md->GetBoundsJ(IndexDomain::entire);
    IndexRange kb = md->GetBoundsK(IndexDomain::entire);
```

src/radiation/imc/imc_driver.cpp
adamdempsey90: This has been added

src/radiation/moments/moments_driver.cpp
adamdempsey90: Fixed

PR #126, opened 2026-04-25
ElCapitan and Kokkos 5 and C++20

adamdempsey90:
## Background

This updates the submodules to the latest versions and puts us on Kokkos 5 and C++20. 

This was needed to get running on El Capitan systems. I've also refreshed / updated the modules/CMake stuff for ElCap.

This requires https://github.com/lanl/jaybenne/pulls 

~I'm still tracking down an issue where coming out of the pgen, the variables are being floored.~ That is now fixed.

~There is what appears to be a memory leak when using GPU aware MPI. Still investigating that.~  This was a missing argument to `flux run`. 

On the ElCap systems, you want to run in CPX mode, e.g. to get 2 nodes:
```
flux alloc -N 2 --amd-gpumode=CPX
```
and then when running artemis you want to use 6 times the number of tasks as GPUs, so in the 2 node example you would do,
```
flux run -N 2 -n 48 --exclusive artemis -i in.par
```
where 48 = 6 * 4 * 2 and `--exclusive` makes sure the task->gpu bindings are correct. 

Running the binary problem with AMR in 2D cylindrical looks good.

## Checklist


CMakeLists.txt
```diff
@@ -89,7 +88,8 @@ if(ARTEMIS_ENABLE_HIP)
   set(JAYBENNE_ENABLE_HIP ON CACHE BOOL "" FORCE)
   set(Kokkos_ENABLE_HIP ON CACHE BOOL "" FORCE)
   set(Kokkos_ENABLE_SERIAL ON CACHE BOOL "" FORCE)
-  set(Kokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE ON CACHE BOOL "" FORCE)
+  set(Kokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE OFF CACHE BOOL "" FORCE)
+  set(Kokkos_ENABLE_IMPL_HIP_MALLOC_ASYNC OFF CACHE BOOL "" FORCE)
```
adamdempsey90, line 92: This was needed to fix a segfault bug deep in kokkos

CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ file(GLOB rebound_src
      "${REBOUND_BUILD_DIR}/src/*.c"
      "${REBOUND_BUILD_DIR}/src/*.h")
 add_custom_command(OUTPUT ${REBOUND_BUILD_DIR}/librebound.so
-                   COMMAND make
+                   COMMAND make CC=${CMAKE_C_COMPILER}
```
adamdempsey90, line 199: Rebound has always just been using `cc` as the C compiler, but when using `amdclang` that was not being set propertly. We probably should've always been doing this though.

CMakePresets.json
```diff
@@ -115,6 +115,72 @@
...
+        "PARTHENON_DISABLE_HDF5_COMPRESSION" : "ON",
+        "HDF5_INCLUDE_DIR": "$env{HDF5_ROOT}/include",
+        "HDF5_INCLUDE_DIRS": "$env{HDF5_ROOT}/include",
+        "HDF5_LIBRARIES": "$env{HDF5_ROOT}/lib",
+        "HDF5_HL_LIBRARIES": "$env{HDF5_ROOT}/lib"
+     }
+    },
+    {
+      "name": "elcap-gpu-debug",
+      "cacheVariables": {
```
adamdempsey90, line 149: GPU debug builds are actually compiling very fast on these systems

adamdempsey90:
Looks like the disk tests are failing on all GPU systems (incl. cuda ones). Must be the update of either parthenon or eos

src/artemis.hpp
pdmullen: Any reason to keep `ARTEMIS_VARIABLE`? Or can we get rid of macro above and just use `PAR_VAR`?

pdmullen (approved):

src/gas/gas.cpp
adamdempsey90: I believe your suspicion was right @pdmullen . Something changed with the singularity units. I have to invert these unit parameters to get the code to run correctly now. I remember @Yurlungur changing something in here recently but I don't see anything in the changelog

adamdempsey90: We also get some of these units from parthenon, maybe something there changed?

adamdempsey90: it looks like this commit inverted the unit we're supposed to pass, https://github.com/lanl/singularity-eos/commit/6ae74f49f968619b234960f0b0e104c4b7e1b179

src/artemis.hpp
adamdempsey90: Using PAR_VAR now

src/gas/gas.cpp
Yurlungur: Yes---this broke backwards compatibility. Sorry for that change.

adamdempsey90: Updated to the latest singularity which included a patch to fix the remaining temperature issue

PR #132, opened 2026-05-26
Draft: More robust restart detection

### LINKED ISSUE #131: Exit codes
- On cuda the signal-triggered shutdown is not cleanly exiting. Throws exit code 143. 

- We should add flip-flop restart dumps to mitigate this issue.

adamdempsey90:
Before, we were checking the actual return code of artemis to determine if we should restart or not. This can run into an issue in slurm where once `srun` detects that one process has exited it will send `SIGTERM` to all remaining processes which causes them to return with code 143 instead of the artemis value. 

Because of this, I have added a file detection based scheme that on startup touches a file called `DO-NOT-RESTART`. The only way to remove this file is for the driver to return with `timeout` at which point the code removes the file. If you try to restart the code and this file is present, it will fatal. Automatic resubmission should check for the existence of that file before restarting. I have updated the docs to reflect this.

I have tested this by hand and it works as expected.

Closes #131 
## Checklist

### DIFF
--- a/doc/src/running.rst
+++ b/doc/src/running.rst
@@ -220,8 +220,7 @@ Return codes
 ^^^^^^^^^^^^
 
 When using batch submissions, it is possible to set up a self-restarting job.
-The easiest way to do this is to take advantage of SLURM interrupt signals and the |code| return code.
-|code|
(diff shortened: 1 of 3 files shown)


pdmullen:
Have you tracked down the failure modes in the regression tests?  Moving to draft until those are resolved.

PR #137, opened 2026-07-14
Improve moments radiation

adamdempsey90:
## Summary

This MR rewrites the M1/P1 radiation–matter coupling source into a globally
convergent implicit solve, makes the moment update's operator splitting an
option (`radiation/moment/split`, defaulting to on), and adds a first-principles
regression test built on the ray-traced disk problem
(`inputs/radiation/raytrace.in`).

The motivating problem is concrete: the ray-traced irradiated disk **does not
run** on `develop`. Its coupling solve stalls (outer iteration never converges)
in the disk's low-density, reduced-speed-of-light cells. The changes here make
that problem run to completion while conserving energy to ~0.2%.

## Motivation — why the old solver stalls on raytrace

The `develop` coupling (`MatterCouplingFullSingleImpl`) is a direct Picard
iteration: it solves the radiation-flux system for `F`, backs out the velocity
change directly (`dv = -icc·dF`), and iterates on `|dEk - dEk_prev|`. That
fixed-point map `F → v → F` has a gain of order

```
mu = eref / (rho * c * chat)
```

The ray-traced disk sits exactly where that gain diverges:

- **Reduced speed of light.** `raytrace.in` uses `creduc = 1000`, so
  `chat = c/1000`. `mu ∝ 1/chat` is therefore ~10³× larger than in a full-`c`
  problem.
- **Low-density cells.** `mu ∝ 1/rho`, and the disk atmosphere spans many
  decades in density. In the tenuous cells the Picard gain exceeds unity, the
  iteration oscillates/diverges, and every such cell runs to `outer_max` every
  step — a stall (or a hard `PARTHENON_FAIL` when `fatal_if_unconverged`).

This is not a tolerance-tuning problem; it is an unstable iteration scheme in
the stiff regime the disk necessarily contains.

### 1. Coupling solver rewrite (`src/radiation/moments/matter_coupling.hpp`)

The direct Picard update is replaced by a local, globally convergent solve. The
pieces, each with its raytrace-centered justification:

- **Solve in the radiation-flux variable, not velocity, in the stiff regime.**
  The momentum step selects its Newton unknown from `mu`
  (`SolveCouplingMomentumEnergy`): flux `F` for `mu ≤ 1`, velocity `beta` for
  `mu > 1`. This is the single most important change for raytrace. When
  `mu ≫ 1`, the flux change implied by a resolvable velocity change is smaller
  than a floating-point ulp of `beta` — `F = Fr0 + (beta0-beta)/mu` barely
  responds to `beta`, so a velocity-variable Newton solve is ill-conditioned and
  cannot converge. Solving in `F` (with `beta = beta0 - mu·delta_F`) is
  well-conditioned across the entire `mu` range.

  *Evidence:* forcing the momentum solve into the velocity variable everywhere
  reproduces the `develop` failure ("Outer not converged") on raytrace, while
  forcing the flux variable everywhere passes. The variable choice is
  load-bearing, independent of the globalization below.

- **Globalized Newton instead of raw Picard.** The local 3×3 system uses a
  finite-difference Jacobian with partial-pivot elimination
  (`SolveCouplingDense3x3`), a backtracking line search, and a stiffness-aware
  fallback direction `1/(1 + (4/3)·mu·E)` that exactly cancels the diffusion-limit
  Picard gain. The old flux-based Picard update (no damping) stalls on raytrace;
  the same flux variable *with* this globalization converges — so the
  globalization is the second necessary ingredient.

- **Ratio-form equilibrium in the optically thick limit**
  (`ComputeCouplingEquilibriumEnergy`). The implicit radiation-energy relation is
  evaluated as `E = E0/(1+ca) + …` rather than by forming `ca·E - cb·B`
  directly. In thick cells `ca, cb` can be O(10¹⁰) and the direct form loses all
  significant digits near LTE. The disk's dense midplane is optically thick, so
  this keeps the inner solve accurate there.

- **Bracketed scalar fallback + conservative energy-only fallback.** When the
  opacity-frozen quasi-Newton inner solve stalls (strongly
  temperature-dependent opacity or a solution pinned on a floor), the inner state
  is found by a bracketed scalar root-find in gas internal energy; if no full
  iterate is admissible, a momentum-frozen, energy-conserving fallback is used
  (also the intentional path for numerical-atmosphere cells). This lets the disk
  atmosphere reradiate absorbed ray energy without applying an unphysical
  radiation drag to floor-density gas.

- **Independent radiation and material floors.** The radiation-energy floor
  (`efloor`) and the material-temperature floor (`tfloor`) are treated as
  independent active constraints, and reference energies are normalized to keep
  numbers near unity. An optically thin cell may legitimately have
  `E_r ≪ a_r T_floor^4`; promoting `E_r` to the blackbody floor created an
  artificial LTE bath and made the constrained system inconsistent on the first
  step.

- **Material temperature-floor correction applied on commit.** The solve seeds
  from the floored internal energy `eg0`, but the commit previously added only
  the solver increment to the *original* conserved state, silently dropping the
  floor lift for cells that started below `tfloor`. The correction
  `eg_floor_delta` is now added to both internal and total gas energy in every
  commit path; it is identically zero when the floor is inactive, so
  zero-opacity problems are bit-for-bit unchanged.

- **No false convergence on a degenerate equilibrium.**
  `ComputeCouplingEquilibriumEnergy` returns failure when the implicit
  radiation-energy linearization is singular (`1 + ca <= 0`, reachable in
  scattering-dominated, mildly relativistic cells). Previously that return was
  discarded, leaving the seed `Eeq = E` so the source residual `E - Eeq`
  collapsed to zero and read as *converged*. All four call sites now propagate
  the failure — the inner scalar residual reports the state inadmissible, the
  quasi-Newton sweep and line-search trials reject it with a large error, and
  the outer recompute clears `solve_valid` — routing the cell to the existing
  bracketed/energy-only fallbacks instead of accepting a non-solution. The
  success path (all current tests, including raytrace) is unchanged.

### 2. Optional operator splitting of the moment update

- New boolean `radiation/moment/split`, **default `true`**.
  - `split = true`: the moment transport + coupling are operator-split from
    hydro and subcycled internally at the radiation CFL, so the global timestep
    is limited by hydro rather than by the (reduced) speed of light.
  - `split = false`: the moment transport + coupling are integrated unsplit
    inside the main RK stages — second-order, no splitting error — at the
    radiation-limited timestep.
- Rationale for defaulting on: with explicit, `chat`-limited moment transport,
  the radiation CFL is typically far more restrictive than the hydro CFL, so
  subcycling avoids dragging the entire timestep down to the radiation limit.
  The unsplit path is retained as the higher-order, splitting-error-free
  reference and for regimes where the two CFLs are comparable.
- Touch points: `artemis_driver.{cpp,hpp}`, `radiation/moments/moments.cpp`
  (registers the mesh timestep estimator only when unsplit), `moments.hpp`,
  `params.yaml`.

### 3. Supporting utilities and cleanup

- `Eps()` and `VNorm()` helpers (`src/utils/artemis_utils.hpp`, `moments.hpp`)
  used throughout the new solver for tolerance scaling and vector norms.
- Removed the unused "simple" matter-coupling path and the dead
  `full_coupling` option (the full solver is the only dispatch target).
- Removed unnecessary finiteness/negativity checks on opacities in the hot loop.

### 4. Ray-trace deposition robustness (`src/radiation/raytrace/raytrace.hpp`)

Numerical guards on the energy deposition: clamp optical depth and path length
to non-negative, use the next term of the `1 - exp(-dtau)` series for small
`dtau`, and correct the inner-boundary extinction path. These prevent spurious
negative or NaN deposition that the coupling solve would otherwise have to
absorb.

### 5. Regression test (`tst/scripts/radiation/raytrace.py`, `raytrace_mpi.py`)

A new test on `inputs/radiation/raytrace.in`, registered in `serial.suite`.
Rather than calibrated "signature" constants, it asserts the conservation law
the solver is built to satisfy, using the injected ray energy as a
first-principles reference:

```
sum_cells vol * [ Δ(E_gas_total) + (c/chat)·Δ(E_rad) ]  ==  sum_cells vol * dt * src_energy
```

- Isolation: hydro and gravity stay on (the disk pgen requires gravity); the
  budget is summed over the whole domain (conservative transport telescopes to a
  small boundary flux) over a single short step, where gravity/coordinate-source
  work is O(dt²) while deposited ray energy is O(dt).
- It also checks the hard invariants the solver must satisfy exactly:
  finiteness, realizability `|F| ≤ E`, subluminal `|v| < c`, and the radiation
  floor, plus a precondition guard that the material floor is initially inactive
  (so the non-conservative floor lift cannot masquerade as an imbalance).
- The problem is run twice on the same input — once split, once unsplit
  (`split=false`) — to cover both integration paths.

## Validation

- **Raytrace now runs and conserves.** Measured global energy-budget residual
  ≈ 2.2×10⁻³ (dominated by boundary flux + O(dt²) gravity work, not the coupling
  law, which converges to the 10⁻¹² outer tolerance); `max|F|/E ≈ 0.32`,
  `max|v|/c ≈ 1.8×10⁻⁴`.
- Both the split and unsplit coupling paths pass the same test.
- The M1 coupling produces the same velocities with and without splitting on
  this problem, consistent with the coupling solve being well-conditioned.

## Behavior / interface changes (reviewer attention)

- **`radiation/moment/split` defaults to `true`.** Existing moment-model inputs
  that do not set it change from unsplit to split integration. In-tree, this
  affects the suite tests `radiation/rad_shock_cgs` and the M1 half of
  `radiation/thermalization`. Both are expected to pass (equilibrium/steady
  states are splitting-independent); `rad_shock_cgs` is the analytic-validated
  one to watch — if it drifts, pin `radiation/moment/split=false` in
  `rad_shock_cgs.in` rather than loosening its L1 tolerance.

### DIFF
--- a/src/artemis_driver.cpp
+++ b/src/artemis_driver.cpp
@@ -89,12 +89,12 @@ ArtemisDriver<GEOM>::ArtemisDriver(ParameterInput *pin, ApplicationInput *app_in
   do_moment_unsplit = false;
   do_moment_split = false;
   if (do_moment) {
-    auto rad_int = pin->GetOrAddString("radiation/moment", "integrator", "rk2");
-    PARTHENON_REQUIRE(((rad_int == "rk1") || (rad_int == "rk2") || (rad_int == "rk3")),
-                      "radiation/integrator must be rk1,rk2, or rk3.")
     do_moment_split = pin->GetOrAddBoolean("radiation/moment", "split", true);
     do_moment_unsplit = !do_moment_split;
     if (do_moment_split) {
+      auto rad_int = pin->GetOrAddString("radiation/moment", "integrator", "rk2");
+      PARTHENON_REQUIRE(((rad_int == "rk1") || (rad_int == "rk2") || (rad_int == "rk3")),
+                        "radiation/integrator must be rk1,rk2, or rk3.")
       rad_integrator = std::make_unique<Integrator_t>(rad_int);
     }
   }
--- a/src/artemis_driver.cpp
+++ b/src/artemis_driver.cpp
@@ -92,7 +92,7 @@ ArtemisDriver<GEOM>::ArtemisDriver(ParameterInput *pin, ApplicationInput *app_in
     auto rad_int = pin->GetOrAddString("radiation/moment", "integrator", "rk2");
     PARTHENON_REQUIRE(((rad_int == "rk1") || (rad_int == "rk2") || (rad_int == "rk3")),
                       "radiation/integrator must be rk1,rk2, or rk3.")
-    do_moment_split = pin->GetOrAddBoolean("radiation/moment", "substep", false);
+    do_moment_split = pin->GetOrAddBoolean("radiation/moment", "split", true);
     do_moment_unsplit = !do_moment_split;
     if (do_moment_split) {
       rad_integrator = std::make_unique<Integrator_t>(rad_int);
--- a/src/artemis_driver.cpp
+++ b/src/artemis_driver.cpp
@@ -83,11 +83,17 @@ ArtemisDriver<GEOM>::ArtemisDriver(ParameterInput *pin, ApplicationInput *app_in
   ndim = pm->ndim;
 
   // Moments integrator
+  do_moment_unsplit = false;
+  do_moment_split = false;
   if (do_moment) {
     auto rad_int = pin->GetOrAddString("radiation/moment", "integrator", "rk2");
     PARTHENON_REQUIRE(((rad_int == "rk1") || (rad_int == "rk2") || (rad_int == "rk3")),
                       "radiation/integrator must be rk1,rk2, or rk3.")
-    rad_integrator = std::make_unique<Integrator_t>(rad_int);
+    do_moment_split = pin->GetOrAddBoolean("radiation/moment", "substep", false);
+    do_moment_unsplit = !do_moment_split;
+    if (do_moment_split) {
+      rad_integrator = std::make_unique<Integrator_t>(rad_int);
+    }
   }
 
   // NBody integrator and initialization
@@ -153,7 +159,7 @@ TaskListStatus ArtemisDriver<GEOM>::Step() {
   }
 
   // Operator split, moments subcycling (M1 or P1)
-  if (do_moment) {
+  if (do_moment_split) {
     status = Moments::MomentsDriver<GEOM>(pmesh, tm, rad_integrator.get());
     if (status != TaskListStatus::complete) return status;
   }
@@ -180,7 +186,7 @@ void ArtemisDriver<GEOM>::PreStepTasks() {
   // set the integration timestep
   integrator->dt = tm.dt;
   if (do_nbody) nbody_integrator->dt = tm.dt;
-  if (do_moment) rad_integrator->dt = tm.dt;
+  if (do_moment_split) rad_integrator->dt = tm.dt;
 
   // Extract Base MeshData Registers
   auto &base = pmesh->mesh_data.Get();
@@ -193,7 +199,7 @@ void ArtemisDriver<GEOM>::PreStepTasks() {
   auto &u1 = pmesh->mesh_data.Add("u1", u0);
 
   // Assign registers with fields required for moments
-  if (do_moment) {
+  if (do_moment_split) {
     parthenon::Metadata::FlagCollection moments_flags, geom_flags;
     moments_flags.TakeUnion(pmesh->packages.Get("moments")->GetMetadataFlag());
     geom_flags.TakeUnion(pmesh->packages.Get("geometry")->GetMetadataFlag());
@@ -258,10 +264,13 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
       // Compute hydrodynamic fluxes
       // NOTE(@adempsey): 1st stage of VL2 uses piecewise constant reconstruction
       const bool do_pcm = ((stage == 1) && (integrator->GetName() == "vl2"));
-      TaskID gas_flx = none, dust_flx = none;
+      TaskID gas_flx = none, dust_flx = none, rad_flx = none;
       if (do_gas && update_fluxes)
         gas_flx = tl.AddTask(none, Gas::CalculateFluxes, u0.get(), do_pcm);
       if (do_dust) dust_flx = tl.AddTask(none, Dust::CalculateFluxes, u0.get(), do_pcm);
+      if (do_moment_unsplit) {
+        rad_flx = tl.AddTask(none, Moments::CalculateFluxes, u0.get());
+      }
 
       // Compute (gas) diffusive fluxes
       TaskID diff_flx = none;
@@ -275,20 +284,23 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
 
       // Communicate and set fluxes
       auto send_flx =
-          tl.AddTask(gas_flx | dust_flx | diff_flx,
+          tl.AddTask(gas_flx | dust_flx | rad_flx | diff_flx,
                      parthenon::SendBoundBufs<parthenon::BoundaryType::flxcor_send>, u0);
       auto recv_flx = tl.AddTask(start_flx_recv, parthenon::ReceiveFluxCorrections, u0);
       auto set_flx = tl.AddTask(recv_flx, parthenon::SetFluxCorrections, u0);
 
       // Apply flux divergence
       auto update =
-          tl.AddTask(gas_flx | dust_flx | set_flx, ArtemisUtils::ApplyUpdate<GEOM>,
-                     u0.get(), u1.get(), g0, g1, bdt);
+          tl.AddTask(gas_flx | dust_flx | rad_flx | set_flx,
+                     ArtemisUtils::ApplyUpdate<GEOM>, u0.get(), u1.get(), g0, g1, bdt);
 
       // Apply "coordinate source terms"
-      TaskID gas_coord_src = update, dust_coord_src = update;
+      TaskID gas_coord_src = update, dust_coord_src = update, rad_coord_src;
       if (do_gas) gas_coord_src = tl.AddTask(update, Gas::FluxSource, u0.get(), bdt);
       if (do_dust) dust_coord_src = tl.AddTask(update, Dust::FluxSource, u0.get(), bdt);
+      if (do_moment_unsplit) {
+        rad_coord_src = tl.AddTask(update, Moments::FluxSource, u0.get(), bdt);
+      }
 
       // Apply (gas) diffusion sources
       TaskID gas_diff_src = gas_coord_src | diff_flx | set_flx;
@@ -311,17 +323,20 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
             tl.AddTask(gravity_src, SelfGravity::SelfGravity<GEOM>, u0.get(), time, bdt);
       }
 
-      TaskID rt_src = self_gravity_src;
+      TaskID rad_src = self_gravity_src;
       // Note that radiation moments will handle this source term if active
-      if (do_raytrace && !do_moment) {
-        rt_src = tl.AddTask(self_gravity_src, Gas::DepositEnergy, u0.get(), bdt);
+      if (do_moment_unsplit) {
+        rad_src =
+            tl.AddTask(self_gravity_src, Moments::MatterCoupling<GEOM>, u0.get(), bdt);
+      } else if (do_raytrace && !do_moment) {
+        rad_src = tl.AddTask(self_gravity_src, Gas::DepositEnergy, u0.get(), bdt);
       }
 
       // Apply rotating frame source term
-      TaskID rframe_src = rt_src;
+      TaskID rframe_src = rad_src;
       if (do_rotating_frame || do_orbital_advection) {
         rframe_src =
-            tl.AddTask(rt_src, RotatingFrame::RotatingFrameForce, u0.get(), time, bdt);
+            tl.AddTask(rad_src, RotatingFrame::RotatingFrameForce, u0.get(), time, bdt);
       }
 
       // Apply drag source term
--- a/src/radiation/moments/matter_coupling.hpp
+++ b/src/radiation/moments/matter_coupling.hpp
@@ -13,6 +13,7 @@
 // the public, perform publicly and display publicly, and to permit others to do
 // so.
 //========================================================================================
+
 // This file was modified in part with the assistance of generative AI.
 #ifndef RADIATION_MOMENTS_MATTER_COUPLING_HPP_
 #define RADIATION_MOMENTS_MATTER_COUPLING_HPP_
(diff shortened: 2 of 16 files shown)


src/radiation/moments/matter_coupling.hpp
```diff
@@ -1,15 +1,20 @@
...
 //
-// This program was produced under U.S. Government contract 89233218CNA000001 for Los
-// Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
-// for the U.S. Department of Energy/National Nuclear Security Administration. All rights
-// in the program are reserved by Triad National Security, LLC, and the U.S. Department
-// of Energy/National Nuclear Security Administration. The Government is granted for
-// itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
-// license in this material to reproduce, prepare derivative works, distribute copies to
-// the public, perform publicly and display publicly, and to permit others to do so.
+// This program was produced under U.S. Government contract 89233218CNA000001
```
pdmullen, line 5: Does your text editor or claude need to be aware of 90 character limit?

src/radiation/moments/matter_coupling.hpp
```diff
@@ -25,180 +30,898 @@ using ArtemisUtils::EOS;
 using ArtemisUtils::MeanOpacity;
 using ArtemisUtils::MeanScattering;
 using ArtemisUtils::VI;
+using ArtemisUtils::VNorm;
 
 namespace Moments {
```
pdmullen, line 35: Just leaving this comment at the top of this...

This solver is now substantially more complex than the previous coupling path. Even if just by private communication (as opposed to an additional CI test), can you show me how this large changeset impacts the solution? Is it avoiding crashes... or providing a higher fidelity solution? Both?

If you really did want to go the route of an added unit test... one where the old solver fails/new solver passes might be worthwhile.

I don’t want to hold this PR up unnecessarily, especially since I understand users need these improvements for active runs. That said, I’m having a hard time understanding which parts of this large change set are actually responsible for the improved behavior.

From the diff, it looks like much of the substantive work is in the matter-coupling solver, so I’m not convinced the improvement is simply from moving moments radiation from an operator-split/subcycled update into the unsplit integrator. Could you help identify the main “difference maker” in this file? 

If the robustness improvement is mostly isolated to one part of the rewrite, I’d like us to consider whether the PR can be narrowed to that piece, or at least documented in a way that makes the necessity of the other changes clear. Right now, without piece-by-piece evidence or a focused reproducer showing which changes improve which failures, I’m being asked to take a fairly large leap of faith that the full change set is needed.

src/radiation/moments/matter_coupling.hpp
```diff
@@ -294,172 +1056,632 @@ TaskStatus MatterCouplingFullSingleImpl(MeshData<Real> *u0, const Real dt) {
...
+            best_dEk = dEk;
+            best_dEr = dEr;
+          }
           if (outer_err <= outer_tol) {
+            outer_converged = true;
             break;
           }
+        } // outer iteration
+
+        // A hard floor turns the nonlinear equality problem into a
```
pdmullen, line 1531: My goodness.... this is starting to feel like HARM's `fixup.cpp`.  How often are we hitting this? Everywhere in atmosphere? Did you ever set a boolean to see where these fixups are hit in practice? 

Or in a realistic problem how bad conservation gets?

src/radiation/moments/moments.hpp
```diff
@@ -43,6 +43,52 @@ TaskCollection MomentsTasks(Mesh *pmesh, const SimTime &tm,
 template <Coordinates GEOM>
 void InitMesh(parthenon::Mesh *pmesh);
 //----------------------------------------------------------------------------------------
+//! \fn Real Moments::EstimateTimeStepMesh
+//! \brief Not enrolled in parthenon's determination for global dt
```
pdmullen: huh? this comment isn't true anymore if we are enrolling moments in an unsplit manner, right? 

On that note---is taking the radiation dt for hydro + other physics packages even a viable path for us? I would have thought that even with reduced speed of light that this would be a showstopper compared to _just_ subcycling the radiatino physics in an operator split update.  

What is the recommendation for users then? Do we want to continue supporting both paths.  It does make our driver a bit more unwieldy.

src/radiation/raytrace/raytrace.cpp
```diff
@@ -242,6 +251,8 @@ TaskStatus CheckCompletion(MeshData<Real> *md) {
 
   auto &rt_pkg = pm->packages.Get("raytrace");
   auto x1max = rt_pkg->Param<Real>("x1max");
+  const Real x1tol =
+      32.0 * std::numeric_limits<Real>::epsilon() * std::max(1.0, std::abs(x1max));
```
pdmullen: is 32 a magic number?

src/radiation/params.yaml
```diff
@@ -45,10 +45,6 @@ radiation:
       _type: Real
       _description: "CFL number for radiation"
       _default: 0.8
-    full_coupling:
```
pdmullen, line 48: does this need to be eliminated from any input files or physics.rst?

src/artemis_driver.cpp
```diff
@@ -83,11 +83,17 @@ ArtemisDriver<GEOM>::ArtemisDriver(ParameterInput *pin, ApplicationInput *app_in
...
   // Moments integrator
+  do_moment_unsplit = false;
+  do_moment_split = false;
   if (do_moment) {
     auto rad_int = pin->GetOrAddString("radiation/moment", "integrator", "rk2");
     PARTHENON_REQUIRE(((rad_int == "rk1") || (rad_int == "rk2") || (rad_int == "rk3")),
                       "radiation/integrator must be rk1,rk2, or rk3.")
-    rad_integrator = std::make_unique<Integrator_t>(rad_int);
+    do_moment_split = pin->GetOrAddBoolean("radiation/moment", "substep", false);
+    do_moment_unsplit = !do_moment_split;
```
pdmullen, line 93: why do we want `rad_int` inside `pin` when `substep==false`? That is confusing because `rad_int != parthenon/time/integrator` is not allowed for unsplit physics.

src/artemis_driver.cpp
```diff
@@ -83,11 +83,17 @@ ArtemisDriver<GEOM>::ArtemisDriver(ParameterInput *pin, ApplicationInput *app_in
   ndim = pm->ndim;
```
pdmullen, line 86: Does everything work correctly with package turn-on and turn-off if using moments unsplit mode? In that event, do we still take the radiation dt?

pdmullen (changes requested):

src/radiation/raytrace/raytrace.hpp
```diff
@@ -130,9 +141,11 @@ TaskStatus PushParticlesImpl(MeshData<Real> *md, const geometry::CoordParams &cp
                 break;
               }
             }
-            dtau *= dx;
+            dtau *= std::max(0.0, dx);
```
RyanWollaeger, line 144: Is the `max` here necessary (can `dx` be 0)?

src/radiation/raytrace/raytrace.hpp
```diff
@@ -109,19 +114,25 @@ TaskStatus PushParticlesImpl(MeshData<Real> *md, const geometry::CoordParams &cp
...
           if (three_d) k = kb.s + inds[2] - ngh;
 
-          while ((i <= ib.e) && (ee > 0.0)) {
+          while ((i >= ib.s) && (i <= ib.e) && (ee > 0.0)) {
             geometry::Coords<GEOM> coords(cpars, pco, k, j, i);
 
             // Deposit energy for this cell and decrement the photon energy
             const auto dx = coords.bnds.x1[1] - coords.bnds.x1[0];
-            Real dtau = vmesh(b, rad::star::absorption(), k, j, i);
+            Real dtau = std::max(0.0, vmesh(b, rad::star::absorption(), k, j, i));
```
RyanWollaeger, line 130: Similar question here - should `rad::star::absorption` be born positivie, making this `max` unneeded? (Sorry if missing a subtlety here.)

src/radiation/moments/matter_coupling.hpp
```diff
@@ -294,172 +1056,632 @@ TaskStatus MatterCouplingFullSingleImpl(MeshData<Real> *u0, const Real dt) {
...
+        v0(b, rad::cons::energy(), k, j, i) = (E0 + dEr) * eref;
         v0(b, gas::cons::momentum(0), k, j, i) += dv[0] * dens * hx[0];
         v0(b, gas::cons::momentum(1), k, j, i) += dv[1] * dens * hx[1];
         v0(b, gas::cons::momentum(2), k, j, i) += dv[2] * dens * hx[2];
-        v0(b, rad::cons::flux(0), k, j, i) += dF[0] * hx[0] * fref;
-        v0(b, rad::cons::flux(1), k, j, i) += dF[1] * hx[1] * fref;
-        v0(b, rad::cons::flux(2), k, j, i) += dF[2] * hx[2] * fref;
+        v0(b, rad::cons::flux(0), k, j, i) = F[0] * hx[0] * fref;
+        v0(b, rad::cons::flux(1), k, j, i) = F[1] * hx[1] * fref;
+        v0(b, rad::cons::flux(2), k, j, i) = F[2] * hx[2] * fref;
```
RyanWollaeger, line 1716: Was this additive stuff the source of the high-velocity problem? (Just curious)

src/radiation/moments/matter_coupling.hpp
RyanWollaeger: Agreed - this is a lot of code to look through - if there is an eagle-eyed view of the critical changes, that would be useful for grokking this.

src/radiation/moments/matter_coupling.hpp
```diff
@@ -25,180 +30,898 @@ using ArtemisUtils::EOS;
...
+  const Real fmag = VNorm(Fin);
+  if (E <= 0.0) return {0.0, 0.0, 0.0};
+  if (fmag <= E || fmag <= Fuzz()) return Fin;
+  const Real scale = E / fmag;
+  return {scale * Fin[0], scale * Fin[1], scale * Fin[2]};
+}
 
-  const auto fatal_if_unconverged =
-      moments_pkg->template Param<bool>("fatal_if_unconverged");
+KOKKOS_INLINE_FUNCTION bool SolveCouplingDense3x3(Real A[3][3], Real rhs[3],
```
RyanWollaeger, line 87: This seems like a generic utility, unless mistaken - could it be moved to `utils` (if only to shorten this matter_coupling file :))?

src/radiation/moments/matter_coupling.hpp
```diff
@@ -25,180 +30,898 @@ using ArtemisUtils::EOS;
...
+          J[row][col] = (residual[row] - Rminus[row]) / h;
+      } else {
+        jacobian_valid = false;
+        break;
+      }
+    }
+
+    std::array<Real, 3> step{0.0, 0.0, 0.0};
+    if (jacobian_valid) {
+      Real rhs[3]{-residual[0], -residual[1], -residual[2]};
```
RyanWollaeger, line 291: Why use C-style array notation here (and in the 3x3 inverter)?

src/radiation/moments/moments.hpp
adamdempsey90: I've backtracked a little on this. I see the same result for split vs. unsplit in the raytracing problem, and the split version gets to the same time faster than unsplit. I've left the user option because there can be times when you want to do unsplit (very coupled). I don't think we'll encounter that in PPD simulations, though.

One additional reason for keeping both paths is that if someone wanted to evaluate the split option for correctness, they would have to run unsplit.

Fixed comment.

src/radiation/moments/matter_coupling.hpp
adamdempsey90: I've added a new test of the raytraced PPD to demonstrate the utility of these changes compared to the original one. The original solver on develop fails immediately with the outer iterations never converging. I tracked this down to the very low density cells stalling on the outer loop to solve for the new flux. The new solver now switches what variable it solves for when mu = eref/(dens c chat) > 1 . If i take this part out of the new solver, then the raytrace test problem again fails, so it seems important.

I've also added a summary of the changes in the description to help understand this whole thing.

But my current feeling with this is: if it works, then we should keep it until it doesn't work. It definitely solved an issue.

src/artemis_driver.cpp
adamdempsey90: Moved this inside the `do_moment_split` guard.

src/radiation/params.yaml
adamdempsey90: It's removed now

src/radiation/raytrace/raytrace.cpp
adamdempsey90: Yes and no, it's a choice of how much off of machine epsilon this tolerance wants to be. I've refactored this a bit so that the numbers are at least named.

pdmullen (approved):

PR #114, opened 2026-01-13
add user source term to pen file "disk.hpp" 

shengtai:
## Background
Following an old PR from Patrick, I have added user source term to "disk.hpp" to reset the region above some scale height in spherical grid. 
 
## Description of Changes
Added a new user parameter "nH_reset". For any grid cell that satisfies nH_reset > 0 and z > nH_reset, the gas velocity is reset to initial conditions, and the dust is reset to zero.
I also added a new parameter in "stokes" regime stopping time so that it is consistent with 2D surface density definition.

## Checklist

- [ ] New features are documented
- [ ] Tests added for bug fixes and new features
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #138, opened 2026-07-23
User source terms

adamdempsey90:
## Background

This adds a simple mechanism for enrolling "user" source tasks into the driver. Disk wave-killing was added as an example.

This supersedes #114 

## Checklist

### DIFF
--- a/src/artemis_driver.cpp
+++ b/src/artemis_driver.cpp
@@ -22,8 +22,8 @@
 
 // Artemis Includes
 #include "artemis.hpp"
-#include "artemis_extras.hpp"
 #include "artemis_driver.hpp"
+#include "artemis_extras.hpp"
 #include "drag/drag.hpp"
 #include "dust/coagulation/coagulation.hpp"
 #include "dust/dust.hpp"
@@ -166,7 +166,6 @@ TaskListStatus ArtemisDriver<GEOM>::Step() {
     status = Dust::Coagulation::CoagulationDriver<GEOM>(pmesh, tm);
     if (status != TaskListStatus::complete) return status;
   }
-
 
   // Other operator split tasks
   for (const auto &tc : GetSplitTaskLists()) {
--- a/src/artemis_driver.cpp
+++ b/src/artemis_driver.cpp
@@ -11,6 +11,8 @@
 // the public, perform publicly and display publicly, and to permit others to do so.
 //========================================================================================
 
+// This file was created in part by generative AI
+
 // NOTE(PDM): The following is largely borrowed from the open-source LANL phoebus
 // software, with additional extensions motivated by other downstream development.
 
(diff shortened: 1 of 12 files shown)


src/artemis_driver.cpp
```diff
@@ -324,11 +326,17 @@ TaskCollection ArtemisDriver<GEOM>::StepTasks() {
             tl.AddTask(rt_src, RotatingFrame::RotatingFrameForce, u0.get(), time, bdt);
       }
 
+      // Apply problem-generator source terms in registration order
+      TaskID user_src = rframe_src;
+      for (const auto &task : GetUserSourceTasks()) {
+        user_src = tl.AddTask(user_src, task.name, task.function, u0.get(), time, bdt);
```
adamdempsey90: I didn't know you could pass names to tasks

src/pgen/problem_modifier.hpp
```diff
@@ -44,6 +46,8 @@ void ProblemModifier(parthenon::ParthenonManager *pman) {
   using BF = parthenon::BoundaryFace;
   using ID = parthenon::IndexDomain;
 
+  ClearUserSourceTasks();
```
pdmullen: Why is this needed? Feels like it could just be eliminated.

pdmullen (approved):
Should the enrollment function name reflect that this is an "unsplit" source term enveloped in the temporal integrator logic --- that is, NOT an operator split source term enrolled after an e.g., `rk2` advance?

adamdempsey90:
> Should the enrollment function name reflect that this is an "unsplit" source term enveloped in the temporal integrator logic --- that is, NOT an operator split source term enrolled after an e.g., `rk2` advance?

I think that clarification is good if we have more than one type of user source term, but if we only have one, then it should be clear from the documentation what it's for (the current documentation is not clear / non-existent). 

I could just add a few of these registries for the different flavors (explicit, implicit, operator-split, etc.)

src/pgen/problem_modifier.hpp
adamdempsey90: Removed

adamdempsey90:
> > Should the enrollment function name reflect that this is an "unsplit" source term enveloped in the temporal integrator logic --- that is, NOT an operator split source term enrolled after an e.g., `rk2` advance?
> 
> I think that clarification is good if we have more than one type of user source term, but if we only have one, then it should be clear from the documentation what it's for (the current documentation is not clear / non-existent).
> 
> I could just add a few of these registries for the different flavors (explicit, implicit, operator-split, etc.)

I added 3 collections, 1 for explicit unsplit tasks, 1 for implicit unsplit tasks ("implicit" in the IMEX sense in terms of ordering), and 1 for split task lists.

